PR #4, opened 2022-08-09
Meta constructor for `BlockDiagLinearOperator` to fix its `matmul` in special case

SebastianAment:
Moving over from [this PR](https://github.com/cornellius-gp/gpytorch/pull/2093).

>BlockDiagLazyTensor's matmul method fails with a dimension mismatch, when BlockDiagLazyTensor is constructed with a DiagLazyTensor as the base_lazy_tensor argument. We hit this case in a BoTorch test, and I added it to the TestBlockDiagLazyTensorMetaClass below.
>
>@Balandat and I thought that intercepting the construction of the BlockDiagLazyTensor and returning a vanilla DiagLazyTensor in this case would be a nice solution, instead of introducing another special case to matmul. This PR implements this feature via the _MetaBlockDiagLazyTensor constructor.

### DIFF
--- a/linear_operator/operators/block_diag_linear_operator.py
+++ b/linear_operator/operators/block_diag_linear_operator.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python3
 
+from abc import ABCMeta
 from typing import Optional, Tuple
 
 import torch
@@ -9,8 +10,26 @@ from ..utils.memoize import cached
 from ._linear_operator import LinearOperator
 from .block_linear_operator import BlockLinearOperator
 
+# metaclass of BlockDiagLinearOperator, overwrites behavior of constructor call
+# _MetaBlockDiagLinearOperator(base_linear_op, block_dim=-3) to return a DiagLazyTensor
+# if base_linear_op is a DiagLinearOperator itself
+class _MetaBlockDiagLinearOperator(ABCMeta):
+    def __call__(cls, base_linear_op, block_dim=-3):
+        from .diag_linear_operator import DiagLinearOperator
+        if cls is BlockDiagLinearOperator and isinstance(base_linear_op, DiagLinearOperator):
+            if not block_dim == -3:
+                raise NotImplementedError(
+                    "Passing a base_linear_op of type DiagLinearOperator to the constructor of "
+                    f"BlockDiagLinearOperator with block_dim = {block_dim} != -3 is not supported."
+                    )
+            else:
+                diag = base_linear_op._diag.flatten(-2, -1) # flatten last two dimensions of diag
+                return DiagLinearOperator(diag)
+        else:
+            return type.__call__(cls, base_linear_op, block_dim)
 
-class BlockDiagLinearOperator(BlockLinearOperator):
+
+class BlockDiagLinearOperator(BlockLinearOperator, metaclass = _MetaBlockDiagLinearOperator):
     """
     Represents a lazy tensor that is the block diagonal of square matrices.
     The :attr:`block_dim` attribute specifies which dimension of the base LinearOperator
--- a/linear_operator/operators/block_diag_linear_operator.py
+++ b/linear_operator/operators/block_diag_linear_operator.py
@@ -10,26 +10,34 @@ from ..utils.memoize import cached
 from ._linear_operator import LinearOperator
 from .block_linear_operator import BlockLinearOperator
 
+
 # metaclass of BlockDiagLinearOperator, overwrites behavior of constructor call
 # _MetaBlockDiagLinearOperator(base_linear_op, block_dim=-3) to return a DiagLazyTensor
 # if base_linear_op is a DiagLinearOperator itself
 class _MetaBlockDiagLinearOperator(ABCMeta):
-    def __call__(cls, base_linear_op, block_dim=-3):
+    def __call__(cls, base_linear_op: LinearOperator, block_dim=-3):
         from .diag_linear_operator import DiagLinearOperator
-        if cls is BlockDiagLinearOperator and isinstance(base_linear_op, DiagLinearOperator):
-            if not block_dim == -3:
+
+        if cls is BlockDiagLinearOperator and isinstance(
+            base_linear_op, DiagLinearOperator
+        ):
+            if block_dim != -3:
                 raise NotImplementedError(
                     "Passing a base_linear_op of type DiagLinearOperator to the constructor of "
                     f"BlockDiagLinearOperator with block_dim = {block_dim} != -3 is not supported."
-                    )
+                )
             else:
-                diag = base_linear_op._diag.flatten(-2, -1) # flatten last two dimensions of diag
+                diag = base_linear_op._diag.flatten(
+                    -2, -1
+                )  # flatten last two dimensions of diag
                 return DiagLinearOperator(diag)
         else:
             return type.__call__(cls, base_linear_op, block_dim)
 
 
-class BlockDiagLinearOperator(BlockLinearOperator, metaclass = _MetaBlockDiagLinearOperator):
+class BlockDiagLinearOperator(
+    BlockLinearOperator, metaclass=_MetaBlockDiagLinearOperator
+):
     """
     Represents a lazy tensor that is the block diagonal of square matrices.
     The :attr:`block_dim` attribute specifies which dimension of the base LinearOperator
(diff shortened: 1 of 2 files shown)


linear_operator/operators/block_diag_linear_operator.py
```diff
@@ -9,8 +10,26 @@
 from ._linear_operator import LinearOperator
 from .block_linear_operator import BlockLinearOperator
 
+# metaclass of BlockDiagLinearOperator, overwrites behavior of constructor call
+# _MetaBlockDiagLinearOperator(base_linear_op, block_dim=-3) to return a DiagLazyTensor
+# if base_linear_op is a DiagLinearOperator itself
+class _MetaBlockDiagLinearOperator(ABCMeta):
+    def __call__(cls, base_linear_op, block_dim=-3):
```
Balandat: Let's add some typing?
```suggestion
    def __call__(cls, base_linear_op: LinearOperator, block_dim: int=-3):
```

linear_operator/operators/block_diag_linear_operator.py
```diff
@@ -9,8 +10,26 @@
...
+        from .diag_linear_operator import DiagLinearOperator
+        if cls is BlockDiagLinearOperator and isinstance(base_linear_op, DiagLinearOperator):
+            if not block_dim == -3:
```
Balandat:
```suggestion
            if block_dim != -3:
```

linear_operator/operators/block_diag_linear_operator.py
```diff
@@ -9,8 +10,26 @@
...
+                    )
+            else:
+                diag = base_linear_op._diag.flatten(-2, -1) # flatten last two dimensions of diag
+                return DiagLinearOperator(diag)
+        else:
+            return type.__call__(cls, base_linear_op, block_dim)
+
 
-class BlockDiagLinearOperator(BlockLinearOperator):
+class BlockDiagLinearOperator(BlockLinearOperator, metaclass = _MetaBlockDiagLinearOperator):
```
Balandat:
```suggestion
class BlockDiagLinearOperator(BlockLinearOperator, metaclass=_MetaBlockDiagLinearOperator):
```

test/operators/test_block_diag_linear_operator.py
```diff
@@ -67,6 +67,36 @@ def evaluate_linear_op(self, linear_op):
...
         return actual
 
+class TestBlockDiagLinearOperatorMetaClass(unittest.TestCase):
+    def test_metaclass_constructor(self):
+        k, n = 3, 5 # number of blocks, block size
+        b1, b2 = 2, 3 # batch dimensions
+        base_operators = [torch.randn(k, n), torch.randn(b1, b2, k, n)]
+
+        for base_op in base_operators: # repeats tests for both batched and non-batched tensors
+            base_diag = DiagLinearOperator(base_op)
```
Balandat: could use [`subTest`](https://docs.python.org/3/library/unittest.html#distinguishing-test-iterations-using-subtests) here to make the test failures easier to parse

test/operators/test_block_diag_linear_operator.py
```diff
@@ -67,6 +67,36 @@ def evaluate_linear_op(self, linear_op):
...
+            linear_op = BlockDiagLinearOperator(base_diag)
+
+            # checks that metaclass constructor returns a diagonal operator
+            # if a DiagLinearOperator is passed to BlockDiagLinearOperator
+            self.assertEqual(type(linear_op), DiagLinearOperator)
+
+            # matrix-vector-multiplication test
+            diag_values = base_op.flatten(start_dim = -2) # operator of non-blocked diagonal values
```
Balandat: Can you make sure to run `black` on this to ensure consistent formatting? (@gpleiss I guess we don't have the commit hooks required for this repo?)
```suggestion
            diag_values = base_op.flatten(start_dim=-2) # operator of non-block diagonal values
```

test/operators/test_block_diag_linear_operator.py
```diff
@@ -67,6 +67,36 @@ def evaluate_linear_op(self, linear_op):
...
+            x = torch.randn(diag_values.shape)
```
Balandat: Suggest using this when possible to ensure dtype and device is inherited from arg
```suggestion
            x = torch.randn_like(diag_values)
```

test/operators/test_block_diag_linear_operator.py
```diff
@@ -67,6 +67,36 @@ def evaluate_linear_op(self, linear_op):
...
+            self.assertTrue(torch.equal(diag_values * x, (linear_op @ x.unsqueeze(-1)).squeeze(-1)))
+
+            # checks that the representation is numerically accurate
+            dense_operator = linear_op.to_dense()
+            truth_operator = torch.diag_embed(diag_values) # creates batch of diagonal operators
+            self.assertTrue(torch.equal(dense_operator, truth_operator))
+
+            with self.assertRaises(NotImplementedError):
```
Balandat: Let's check that this spits out the right error message
```suggestion
            with self.assertRaisesRegex(
                NotImplementedError,
                "with block_dim = -2 != -3 is not supported"   
            ):
```

Balandat:
hmm @gpleiss I somehow can't "Approve and run" the tests on this PR. Any idea why this may be?

Balandat:
> hmm @gpleiss I somehow can't "Approve and run" the tests on this PR. Any idea why this may be?

Nevermind, looks like it was just taking some time to get going.

Balandat:
> > hmm @gpleiss I somehow can't "Approve and run" the tests on this PR. Any idea why this may be?
> 
> Nevermind, looks like it was just taking some time to get going.

Actually now it doesn't show anymore. Weird.

linear_operator/operators/block_diag_linear_operator.py
```diff
@@ -10,7 +11,33 @@
...
+        ):
+            if block_dim != -3:
+                raise NotImplementedError(
+                    "Passing a base_linear_op of type DiagLinearOperator to the constructor of "
+                    f"BlockDiagLinearOperator with block_dim = {block_dim} != -3 is not supported."
+                )
+            else:
+                diag = base_linear_op._diag.flatten(
+                    -2, -1
+                )  # flatten last two dimensions of diag
```
Balandat: readability nit
```suggestion
                # flatten last two dimensions of diag
                diag = base_linear_op._diag.flatten(-2, -1)  
```

linear_operator/operators/block_diag_linear_operator.py
```diff
@@ -56,16 +83,22 @@ def _diagonal(self):
 
     def _get_indices(self, row_index, col_index, *batch_indices):
         # Figure out what block the row/column indices belong to
-        row_index_block = torch.div(row_index, self.base_linear_op.size(-2), rounding_mode="floor")
-        col_index_block = torch.div(col_index, self.base_linear_op.size(-1), rounding_mode="floor")
+        row_index_block = torch.div(
```
Balandat: Looks like black changed this? Probably b/c it used the standard 88 line length rather than 120 (which I think has been used across this codebase?). Let's reformat with passing `--line-length 120` to black to make sure this PR doesn't change a bunch of things unrelated to it functionality.

Balandat (approved):
Another formatting nit about line lengths, o/w this lgtm

SebastianAment:
Thank you for the suggestions! I incorporated them and ran `black` over the two files, which also made changes to lines I hadn't touched before. Curiously, `black` actually fails in the build step here on GitHub with the new version. Maybe it has different settings locally?

Balandat (approved):
thanks!

Balandat:
Actually, looks like `isort` is still complaining about the import order. Once that's fixed this can go in

SebastianAment:
If I am parsing the output correctly, it doesn't seem to be happy about `test_block_diag_linear_operator.py`. When I run `isort --line-length 120` on this file locally, it just gets rid of the parentheses for the import from `linear_operator.operators` but doesn't change the order.
Do you know how we can determine what specifically it isn't happy about?

Balandat:
Hmm not sure. Are you running the same isort version & settings as defined in the pre-commit config? https://github.com/cornellius-gp/linear_operator/blob/main/.pre-commit-config.yaml#L27-L32

 Maybe it wants to sort the items of the import alphabetically?

SebastianAment:
I ran `isort -w120 -m3 --tc --project=gpytorch .` with `isort` version 5.10.1, the same as this repo. The only change I get is the deleted parentheses in `test_block_diag_linear_operator.py` and a re-ordering of imports in `docs/source/conf.py`. Since the latter is excluded from `isort` on here, I am not committing that change. Amending and force-pushing b/c there are no meaningful differences to the last version.

PR #9, opened 2022-09-05
Update _linear_operator.py

### LINKED ISSUE #8: Pythonic Sum is broken with LinearOperators
Attempting to use Pythonic sum of arbitrary iterables of LinearOperators raises TypeError. This is because the builtin sum() function starts with int(0) and sums iteratively, as follows:

```
sum([a,b,..z]) = 0 + a + b + .. + z
```

## Minimal reproducible example (with workaround)

```
import gpytorch
import torch
import traceback
import warnings

x1 = torch.rand([16,1])
x2 = torch.rand([16,1])

k1 = gpytorch.kernels.RBFKernel()(x1)
k2 = gpytorch.kernels.RBFKernel()(x2)

k3 = k1+k2

print("'+' operator works fine.")

# This does not
try:
    k4 = sum([k1,k2])
except Exception as e: traceback.print_exc()

# A workaround now: functools.reduce works, but is a little unsightly

from functools import reduce
import operator

k5 = reduce(operator.add,[k1,k2])
```

## A Fix

This can be fixed by simply including logic for LinearOperator to add int(0) in the same way as it adds a ZeroLinearOperator, by returning ```self```, as follows:

```
import numbers
# Specifically only allow numbers.Number without raising TypeError in the case where other==0
if isinstance(other, numbers.Number) and other==0:
    return self
```

I'd recommend adding this case at the end of the conditions in LinearOperator.\_\_add\_\_, on line 2571, so it's only checked after the more common usecases, as follows:

```
def __add__(self, other: Union[torch.Tensor, LinearOperator, float]) -> LinearOperator:
    from torch import Tensor

    from .added_diag_linear_operator import AddedDiagLinearOperator
    from .dense_linear_operator import to_linear_operator
    from .diag_linear_operator import DiagLinearOperator
    from .root_linear_operator import RootLinearOperator
    from .sum_linear_operator import SumLinearOperator
    from .zero_linear_operator import ZeroLinearOperator
    
    import numbers.Number

    if isinstance(other, ZeroLinearOperator):
        return self
    elif isinstance(other, DiagLinearOperator):
        return AddedDiagLinearOperator(self, other)
    elif isinstance(other, RootLinearOperator):
        return self.add_low_rank(other.root)
    elif isinstance(other, Tensor):
        other = to_linear_operator(other)
        shape = torch.broadcast_shapes(self.shape, other.shape)
        new_self = self if self.shape[:-2] == shape[:-2] else self._expand_batch(shape[:-2])
        new_other = other if other.shape[:-2] == shape[:-2] else other._expand_batch(shape[:-2])
        return SumLinearOperator(new_self, new_other)
    elif isinstance(other, numbers.Number) and other==0:
        return self
    else:
        return SumLinearOperator(self, other)

```

wbeardall:
Add additional LinearOperator.__add__ case for Pythonic addition operations with the builtin sum() function. After other common use-cases, check if other is a numeric Python type and equal to zero, and if so, return self.

Fixes #8.

### DIFF
--- a/linear_operator/operators/_linear_operator.py
+++ b/linear_operator/operators/_linear_operator.py
@@ -2556,6 +2556,8 @@ class LinearOperator(ABC):
         from .sum_linear_operator import SumLinearOperator
         from .zero_linear_operator import ZeroLinearOperator
 
+        import numbers
+
         if isinstance(other, ZeroLinearOperator):
             return self
         elif isinstance(other, DiagLinearOperator):
@@ -2568,6 +2570,8 @@ class LinearOperator(ABC):
             new_self = self if self.shape[:-2] == shape[:-2] else self._expand_batch(shape[:-2])
             new_other = other if other.shape[:-2] == shape[:-2] else other._expand_batch(shape[:-2])
             return SumLinearOperator(new_self, new_other)
+        elif isinstance(other,numbers.Number) and other==0:
+            return self
         else:
             return SumLinearOperator(self, other)
 
--- a/linear_operator/operators/_linear_operator.py
+++ b/linear_operator/operators/_linear_operator.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import functools
 import math
+import numbers
 import warnings
 from abc import ABC, abstractmethod
 from copy import deepcopy
@@ -2546,17 +2547,13 @@ class LinearOperator(ABC):
 
         return samples
 
-    def __add__(self, other: Union[torch.Tensor, LinearOperator, float]) -> LinearOperator:
-        from torch import Tensor
-
+    def __add__(self, other: Union[Tensor, LinearOperator, float]) -> LinearOperator:
         from .added_diag_linear_operator import AddedDiagLinearOperator
         from .dense_linear_operator import to_linear_operator
         from .diag_linear_operator import DiagLinearOperator
         from .root_linear_operator import RootLinearOperator
         from .sum_linear_operator import SumLinearOperator
         from .zero_linear_operator import ZeroLinearOperator
-
-        import numbers
 
         if isinstance(other, ZeroLinearOperator):
             return self
--- a/linear_operator/operators/_linear_operator.py
+++ b/linear_operator/operators/_linear_operator.py
@@ -2570,7 +2570,7 @@ class LinearOperator(ABC):
             new_self = self if self.shape[:-2] == shape[:-2] else self._expand_batch(shape[:-2])
             new_other = other if other.shape[:-2] == shape[:-2] else other._expand_batch(shape[:-2])
             return SumLinearOperator(new_self, new_other)
-        elif isinstance(other,numbers.Number) and other==0:
+        elif isinstance(other, numbers.Number) and other == 0:
             return self
         else:
             return SumLinearOperator(self, other)


linear_operator/operators/_linear_operator.py
```diff
@@ -2568,6 +2570,8 @@ def __add__(self, other: Union[torch.Tensor, LinearOperator, float]) -> LinearOp
             new_self = self if self.shape[:-2] == shape[:-2] else self._expand_batch(shape[:-2])
             new_other = other if other.shape[:-2] == shape[:-2] else other._expand_batch(shape[:-2])
             return SumLinearOperator(new_self, new_other)
+        elif isinstance(other,numbers.Number) and other==0:
```
Balandat:
```suggestion
        elif isinstance(other, numbers.Number) and other == 0:
```

Balandat (approved):
Thanks for the fix, a formatting nit

linear_operator/operators/_linear_operator.py
```diff
@@ -2556,6 +2556,8 @@ def __add__(self, other: Union[torch.Tensor, LinearOperator, float]) -> LinearOp
         from .sum_linear_operator import SumLinearOperator
         from .zero_linear_operator import ZeroLinearOperator
 
+        import numbers
+
         if isinstance(other, ZeroLinearOperator):
             return self
```
SebastianAment: This is tangential to this PR but touches the same function:

Shouldn't we return a copy of `self` in the `ZeroLinearOperator` branch? 

Generally, out-of-place addition yields a new object in memory. A user could therefore assume that the returned sum is independent of the summands. This seems especially relevant to `DenseLinearOperator`, whose [add method](https://github.com/cornellius-gp/linear_operator/blob/2e66c243a8ffc4359512bc270cec3f796a304688/linear_operator/operators/dense_linear_operator.py#L79) falls back to this generic implementation, but similarly to `DiagLinearOperator` etc.

E.g. assigning a new value to some of the entries of the returned sum
would cause unintended side-effects.

Balandat: Hmm, that seems like a good point to me. In general I think we should try to not assign new values to entries of `LinearOperator` objects post construction, but I guess that could happen so better be safe than sorry (especially since returning a new `ZeroLinearOperator` will be cheap). Curious what @gpleiss's thoughts are on this.

SebastianAment: It might be more expensive since `self` is generally not a `ZeroLinearOperator`, even though `other` is. 

Also, a few operators have the `base_linear_op` field, which signals that it is not a private variable, and could be up for mutation.

I introduced [the `copy` here](https://github.com/cornellius-gp/linear_operator/pull/10). We can continue the discussion there as to not block @wbeardall in merging this PR.

Balandat: > It might be more expensive since self is generally not a ZeroLinearOperator, even though other is.

D'oh. Of course. 

> Also, a few operators have the base_linear_op field, which signals that it is not a private variable, and could be up for mutation.

Hmm I think in general one should not mutate the `base_linear_op` field outside of the `LinearOperator`'s methods. If that's universally accepted we may want to rename and `privatize` the constituent operators.

gpleiss: > Hmm I think in general one should not mutate the base_linear_op field outside of the LinearOperator's methods. If that's universally accepted we may want to rename and privatize the constituent operators.

Agreed 👍

linear_operator/operators/_linear_operator.py
```diff
@@ -2568,6 +2570,8 @@ def __add__(self, other: Union[torch.Tensor, LinearOperator, float]) -> LinearOp
             new_self = self if self.shape[:-2] == shape[:-2] else self._expand_batch(shape[:-2])
             new_other = other if other.shape[:-2] == shape[:-2] else other._expand_batch(shape[:-2])
             return SumLinearOperator(new_self, new_other)
+        elif isinstance(other, numbers.Number) and other == 0:
+            return self
```
gpleiss, line 2574:
```suggestion
            return deepcopy(self)
```

As per the discussion above.

gpleiss (changes requested):
We should perform a deep copy of self (see discussion).

SebastianAment:
@gpleiss We actually had to [back out the 'deepcopy'](https://github.com/cornellius-gp/linear_operator/pull/12) again because it broke things downstream in Botorch. I'd say let's not include that here until we understand how to fix it.

gpleiss:
@SebastianAment 😢 

Okay, let's merge this PR as is (well... after fixing linting), and then I'll open an issue so that we don't forget to switch this to deep copy one day.

gpleiss (approved):

PR #10, opened 2022-09-07
Edits to generic `add`, `BlockDiagLinearOperator`'s `matmul`, and documentation

SebastianAment:
This PR makes three small changes:

1) It introduces a size check in `BlockDiagLinearOperator`'s `matmul`. This takes care of an edge case where the *blocks* of two block diagonal operators have different sizes, even though the operators have the same size. E.g. One operator has twice as many blocks of half the size as the other.

2) It introduces a `deepcopy` in the generic `__add__` method for addition with a `ZeroLinearOperator`, see [the discussion here](https://github.com/cornellius-gp/linear_operator/pull/9#discussion_r964924759).

3) It makes edits to the docs, which referenced the old `lazy` submodule, instead of the new `operators`.

### DIFF
--- a/linear_operator/test/linear_operator_test_case.py
+++ b/linear_operator/test/linear_operator_test_case.py
@@ -883,13 +883,13 @@ class LinearOperatorTestCase(RectangularLinearOperatorTestCase):
             linear_op = self.create_linear_op()
             rootdecomp = linear_operator.root_decomposition(linear_op)
 
-            if isinstance(rootdecomp, linear_operator.lazy.CholLinearOperator):
+            if isinstance(rootdecomp, linear_operator.operators.CholLinearOperator):
                 chol = linear_operator.root_decomposition(linear_op).root.clone()
                 linear_operator.utils.memoize.clear_cache_hook(linear_op)
                 linear_operator.utils.memoize.add_to_cache(
                     linear_op,
                     "root_decomposition",
-                    linear_operator.lazy.RootLinearOperator(chol),
+                    linear_operator.operators.RootLinearOperator(chol),
                 )
 
                 _wrapped_cholesky = MagicMock(wraps=torch.linalg.cholesky_ex)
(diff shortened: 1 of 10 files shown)


SebastianAment:
While thinking about the sizes in `matmul`, I noticed that the docs of `BlockLinearOperator` imply that the blocks should be square. If we don't want to support rectangular blocks, I'd also add a quick check in the constructor of `BlockLinearOperator` to make sure the blocks are indeed square.

linear_operator/test/linear_operator_test_case.py
```diff
@@ -883,13 +883,13 @@ def _test_triangular_linear_op_inv_quad_logdet(self):
             linear_op = self.create_linear_op()
             rootdecomp = linear_operator.root_decomposition(linear_op)
 
-            if isinstance(rootdecomp, linear_operator.lazy.CholLinearOperator):
```
Balandat, line 886: Huh. Did this test never run? This should have failed, no?

Balandat (approved):
lgtm, thanks for fixing the docs. Let's add that check for blocks being square as well (that has so far been an implicit assumption for most of this library anyway, though ideally we can relax this going forward).

linear_operator/test/linear_operator_test_case.py
SebastianAment: It seems to never have been executed because the method starts with an underscore ("\_test\_..."). Removing the underscore makes this execute, but fail.

Balandat: Ugh. OK how about we land this as is (with underscore) and create an issue for fixing the test in a separate PR? (unless this is a quick fix).

Balandat: tests-cant-fail-5b5b40

SebastianAment: I'll spend a few more minutes on the test to see if I can figure it out quickly. It puzzles me at the moment.

SebastianAment: I understand the failures now, but did not fix it in this PR. Raised an [issue here](https://github.com/cornellius-gp/linear_operator/issues/11).

PR #14, opened 2022-09-13
Performant Scaling of `BlockDiagLinearOperator` by `DiagLinearOperator`

SebastianAment:
The primary goal of this PR is to enable the efficient scaling of `BlockDiagLinearOperators` by `DiagLinearOperators`. This will allow us to remove a special case in [BoTorch's outcome transform](https://github.com/pytorch/botorch/blob/5f2028d8f6883eed29fe789bbc54f240937f6125/botorch/models/transforms/outcome.py#L389). 

In order to achieve this, this PR modifies and adds special cases to `DiagLinearOperator` and `BlockDiagLinearOperator`'s `matmul`.

I tested the notebook that exhibited the regression by patching in the function definitions here and ended up with a 16 second runtime - as opposed to 30+ minutes before.

### DIFF
--- a/linear_operator/operators/_linear_operator.py
+++ b/linear_operator/operators/_linear_operator.py
@@ -558,12 +558,7 @@ class LinearOperator(ABC):
         if isinstance(self, DenseLinearOperator) or isinstance(other, DenseLinearOperator):
             return DenseLinearOperator(self.to_dense() * other.to_dense())
         else:
-            left_linear_op = self if self._root_decomposition_size() < other._root_decomposition_size() else other
-            right_linear_op = other if left_linear_op is self else self
-            return MulLinearOperator(
-                left_linear_op.root_decomposition(),
-                right_linear_op.root_decomposition(),
-            )
+            return MulLinearOperator(self, other)
 
     def _preconditioner(self) -> Tuple[Callable, "LinearOperator", torch.Tensor]:
         """
--- a/linear_operator/operators/block_diag_linear_operator.py
+++ b/linear_operator/operators/block_diag_linear_operator.py
@@ -153,8 +153,9 @@ class BlockDiagLinearOperator(BlockLinearOperator, metaclass=_MetaBlockDiagLinea
             return BlockDiagLinearOperator(self.base_linear_op @ other.base_linear_op)
         # special case if we have a DiagLinearOperator
         if isinstance(other, DiagLinearOperator):
-            diag_reshape = other._diag.view(*self.base_linear_op.shape[:-2], 1, -1)
-            return BlockDiagLinearOperator(self.base_linear_op * diag_reshape)
+            diag_reshape = other._diag.view(*self.base_linear_op.shape[:-1])
+            diag = DiagLinearOperator(diag_reshape)
+            return BlockDiagLinearOperator(self.base_linear_op @ diag)
         return super().matmul(other)
 
     @cached(name="svd")
--- a/linear_operator/operators/block_diag_linear_operator.py
+++ b/linear_operator/operators/block_diag_linear_operator.py
@@ -153,6 +153,7 @@ class BlockDiagLinearOperator(BlockLinearOperator, metaclass=_MetaBlockDiagLinea
             return BlockDiagLinearOperator(self.base_linear_op @ other.base_linear_op)
         # special case if we have a DiagLinearOperator
         if isinstance(other, DiagLinearOperator):
+            # matmul is going to be cheap because of the special casing in DiagLinearOperator
             diag_reshape = other._diag.view(*self.base_linear_op.shape[:-1])
             diag = DiagLinearOperator(diag_reshape)
             return BlockDiagLinearOperator(self.base_linear_op @ diag)
--- a/linear_operator/operators/dense_linear_operator.py
+++ b/linear_operator/operators/dense_linear_operator.py
@@ -23,7 +23,7 @@ class DenseLinearOperator(LinearOperator):
         Args:
         - tsr (Tensor: matrix) a Tensor
         """
-        super(DenseLinearOperator, self).__init__(tsr)
+        super().__init__(tsr)
         self.tensor = tsr
 
     def _cholesky_solve(self, rhs, upper=False):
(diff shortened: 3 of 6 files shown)


linear_operator/operators/_linear_operator.py
```diff
@@ -558,12 +558,7 @@ def _mul_matrix(self, other: Union[torch.Tensor, "LinearOperator"]) -> LinearOpe
         if isinstance(self, DenseLinearOperator) or isinstance(other, DenseLinearOperator):
```
SebastianAment, line 559: This line was the source of the performance regressions. The reasoning behind it appears to be because `MulLinearOperator` *always* computes a root decomposition, which is both inefficient and introduces dead code in its implementation (see below). I am sidestepping this by replacing the `*` with secondary `@` operators in the new special cases of the `DiagLinearOperator` and `BlockDiagLinearOperator` `matmul` methods, leading to `MatmulLinearOperators` instead.

However, this does not get rid of the more general issue. To fix that, I propose two steps in a future PR:
1) Introducing logic in the constructor of `MulLinearOperator` that decides whether or not to build a root decomposition.
2) Even if a root decomposition seems beneficial, delaying its computation until the very last moment when it is needed in `matmul` and cache the result. This will give us ~0 overhead in the case where the linear operator represents a posterior covariance matrix that is constructed via a `posterior` call but only the posterior mean is needed, as was the case in the notebook that exhibited the regression.

linear_operator/operators/dense_linear_operator.py
```diff
@@ -76,13 +76,7 @@ def __add__(self, other):
         elif isinstance(other, torch.Tensor):
             return DenseLinearOperator(self.tensor + other)
         else:
-            return super(DenseLinearOperator, self).__add__(other)
-
-    def mul(self, other):
```
SebastianAment, line 81: As long as the case for `DenseLinearOperator` exists in `_mul_matrix` in `_linear_operator.py`, this is not necessary.

linear_operator/operators/block_diag_linear_operator.py
```diff
@@ -153,8 +153,9 @@ def matmul(self, other):
             return BlockDiagLinearOperator(self.base_linear_op @ other.base_linear_op)
         # special case if we have a DiagLinearOperator
         if isinstance(other, DiagLinearOperator):
-            diag_reshape = other._diag.view(*self.base_linear_op.shape[:-2], 1, -1)
-            return BlockDiagLinearOperator(self.base_linear_op * diag_reshape)
+            diag_reshape = other._diag.view(*self.base_linear_op.shape[:-1])
+            diag = DiagLinearOperator(diag_reshape)
+            return BlockDiagLinearOperator(self.base_linear_op @ diag)
```
Balandat, line 159: may be worth adding a comment here that this matmul is going to be cheap b/c of the special casing in `DiagLinearOperator`?

linear_operator/operators/diag_linear_operator.py
```diff
@@ -164,20 +154,29 @@ def log(self) -> "DiagLinearOperator":
...
-        if isinstance(other, DenseLinearOperator):
-            return DenseLinearOperator(self._diag.unsqueeze(-1) * other.tensor)
-        # special case if we have a BlockDiagLinearOperator
-        if isinstance(other, BlockDiagLinearOperator):
-            diag_reshape = self._diag.view(*other.base_linear_op.shape[:-1], 1)
-            return BlockDiagLinearOperator(diag_reshape * other.base_linear_op)
+        elif isinstance(other, BlockDiagLinearOperator):
+            diag_reshape = self._diag.view(*other.base_linear_op.shape[:-1])
+            diag = DiagLinearOperator(diag_reshape)
+            # using matmul here avoids having to implement special case of elementwise multiplication
```
Balandat, line 162: should we implement that special case of elementwise multiplication either way though?

linear_operator/operators/mul_linear_operator.py
```diff
@@ -62,7 +65,7 @@ def _matmul(self, rhs):
             left_res = left_res.view(*output_batch_shape, n, rank, m)
             res = left_res.mul_(left_root.unsqueeze(-1)).sum(-2)
         # This is the case where we're not doing a root decomposition, because the matrix is too small
-        else:
+        else:  # Dead?
```
Balandat, line 68: This is b/c the `MulLinearOperator` always constructs a root decomposition?

Balandat (approved):
Overall this lgtm but I'll let @gpleiss double check the broader implications of this change.

linear_operator/operators/mul_linear_operator.py
SebastianAment: Yes, exactly.

linear_operator/operators/diag_linear_operator.py
SebastianAment: Yes, it'd make the code more robust to other use cases.

This was actually the first thing I tried to solve the regression, but it would have been a more complex solution for the performance issue we ran into. This is because for multiplication with a diagonal, we only need to support a broadcasted element-wise multiplication *with a vector*, and not the general case where it could also be a higher-order tensor. This complicates things, especially if we multiply with a structured operator like a `BlockDiagLinearOperator`. 

The specific implementation of the element-wise multiplication will depend on what we do [regarding `_mul_matrix`](https://github.com/cornellius-gp/linear_operator/pull/14/files/0c64872ce155b3e8e2c2880fb595eadd986e89d5#r969977789) and is probably best deferred to that future PR.

linear_operator/operators/block_diag_linear_operator.py
```diff
@@ -153,8 +153,9 @@ def matmul(self, other):
             return BlockDiagLinearOperator(self.base_linear_op @ other.base_linear_op)
         # special case if we have a DiagLinearOperator
         if isinstance(other, DiagLinearOperator):
-            diag_reshape = other._diag.view(*self.base_linear_op.shape[:-2], 1, -1)
-            return BlockDiagLinearOperator(self.base_linear_op * diag_reshape)
+            diag_reshape = other._diag.view(*self.base_linear_op.shape[:-1])
```
gpleiss, line 157:
```suggestion
            # matmul is going to be cheap because of the special casing in DiagLinearOperator
            diag_reshape = other._diag.view(*self.base_linear_op.shape[:-1])
```

linear_operator/operators/_linear_operator.py
gpleiss: I think this change should be fine, since the MulLinearOpeator constructor performs root decompositions on left_linear_op and right_linear_op: https://github.com/cornellius-gp/linear_operator/blob/caead4defb7b1ccd3728f3fbff5bc4c63aa0b753/linear_operator/operators/mul_linear_operator.py#L25

> Even if a root decomposition seems beneficial, delaying its computation until the very last moment when it is needed in matmul and cache the result.

Agreed.

gpleiss (approved):

PR #15, opened 2022-09-13
Removing `evaluate_kernel`

SebastianAment:
This PR aims at removing `evaluate_kernel` from the `linear_operator` package. This was already suggested in a comment in the code and makes sense since -- as far as I can tell -- it is a no-op for all operators in this package.

The main question at this point seems to be if additional changes are required in GPyTorch, which defines `LazyEvaluatedKernelTensor` and should therefore take care of any of its implementation specifics.

### DIFF
--- a/linear_operator/operators/_linear_operator.py
+++ b/linear_operator/operators/_linear_operator.py
@@ -469,12 +469,10 @@ class LinearOperator(ABC):
         from .keops_linear_operator import KeOpsLinearOperator
         from .triangular_linear_operator import TriangularLinearOperator
 
-        evaluated_kern_mat = self.evaluate_kernel()
-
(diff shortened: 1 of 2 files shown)


Balandat (commented):
There are a number of call sites in gpytorch that @gpleiss or @jacobrgardner have a much better understanding of, so I'll defer to their input on this.

Balandat:
As to the changes here, they look good in isolation.

jacobrgardner:
I just want to be careful about making sure these don't get used. The problem these lines of code originally solved is that, with the current lazy kernel evaluation design, ALL covariance matrices are LazyEvaluatedKernelTensors until we evaluate the kernel, so if we need to type check the kind of linear operator we get out of a Kernel call we can't do that until the kernel has been evaluated

gpleiss:
It would be great if we could remove this, but I agree with Jake that we want to be extra careful :)

PR #18, opened 2022-09-26
Update Readme

gpleiss:


README.md
```diff
@@ -1,12 +1,332 @@
...
+            torch.linalg.cholesky(eye + C.mT @ D_invC, upper=False),
+            upper=False
+        )
+    )
+
+
+# C = torch.randn(1000, 20)
+# d = torch.randn(1000)
+# b = torch.randn(1000)
+low_rank_plus_diagonal_solve(C, d, b)  # computes A^{-1} b efficiently
```
Balandat: Should we say how much more efficiently or is this too much in the weeds?

README.md
```diff
@@ -1,12 +1,332 @@
...
+```python
+from linear_operator.operators import DiagLinearOperator, RootLinearOperator
+# C = torch.randn(1000, 20)
+# d = torch.randn(1000)
+# b = torch.randn(1000)
+A = RootLinearOperator(C) + DiagLinearOperator(d)  # represents C C^T + diag(d)
+```
+
+it provides an interface that lets us treat $\boldsymbol A$ as if it were a generic tensor,
+using the standard PyTorch API...
```
Balandat:
```suggestion
using the standard PyTorch API:
```

README.md
```diff
@@ -1,12 +1,332 @@
...
+
+```python
+torch.linalg.solve(A, d, b)  # computes A^{-1} b efficiently!
 ```
-python -m unittest discover
+
+but under-the-hood, the `LinearOperator` object keeps track of the algebraic structure of $\boldsymbol A$ (low rank plus diagonal)
```
Balandat:
```suggestion
Under-the-hood, the `LinearOperator` object keeps track of the algebraic structure of $\boldsymbol A$ (low rank plus diagonal)
```

README.md
```diff
@@ -1,12 +1,332 @@
...
+and determines the most efficient routine to use (the Woodbury formula).
+This way, we can get a efficient ($\mathcal O(N)$) solve while abstracting away all of the details.
```
Balandat: this doesn't seem to render properly, probably b/c of the outer parentheses?

README.md
```diff
@@ -1,12 +1,332 @@
...
+D = DiagLinearOperator(torch.tensor([1., 2., 3.])
+# Represents the matrix
+#   [[1., 0., 0.],
+#    [0., 2., 0.],
+#    [0., 0., 3.]]
+torch.matmul(D, torch.tensor([4., 5., 6.])
+# Returns [4., 10., 18.]
+```
+
+Other functions (e.g. `logdet`, `eigh`, etc.) can be defined as needed if efficient routines exist.
```
Balandat, line 170: We should make this more prominent, this is easily skipped in the current layout in which case the package seems a lot less impressive / useful.

README.md
```diff
@@ -1,12 +1,332 @@
...
+class DiagLinearOperator(linear_operator.LinearOperator):
+    r"""
+    A LinearOperator representing a diagonal matrix.
+    """
+    def __init__(self, diag):
+        # diag: the vector that defines the diagonal of the matrix
+        self.diag = diag
+
+    def _matmul(self):
+        return torch.Size([*self.diag.shape, self.diag.size(-1)])
```
Balandat: This is wrong (C&P issue it seems)

README.md
```diff
@@ -1,12 +1,332 @@
...
+Other functions (e.g. `logdet`, `eigh`, etc.) can be defined as needed if efficient routines exist.
+
+#### Why is This Useful?
+While `_matmul`, `_size`, and `_transpose_nonbatch` might seem like a limited set of functions,
+it turns out that most functions on the `torch` and `torch.linalg` namespaces can be efficiently implemented
+using only these three primitative functions.
+
+Moreover, because `_matmul` is a linear function, it is very easy to compose linear operators in various ways.
+For example: adding two linear operators (`SumLinearOperator`) just requires adding the output of their `_matmul` functions.
+This makes it possible to define very complex compositional structures that still yield efficient linear algebraic routines.
```
Balandat, line 179: We should also talk about the fact that the product (not in the mathematical sense) of an operation on one or more `LinearOperator` objects can again return a `LinearOperator`, i.e. that the operations are structure preserving. E.g. we can give the example of summing two `DiagLinearOperator`s.

README.md
```diff
@@ -1,12 +1,332 @@
...
+from gpytorch.distributions import MultivariateNormal
+
+# variance = torch.randn(10000)
+cov = DiagLinearOperator(variance)
+# or
+# cov = RootLinearOperator(...) + DiagLinearOperator(...)
+# or
+# cov = KroneckerProductLinearOperator(...)
+# or
+# cov = ToeplitzLinearOperator(...)
```
Balandat: So I played around with this, and unfortunately this doesn't work as well as we'd like (yet). E.g. the low rank + diag case will in the `MultivariateNormal` code still perform a Cholesky decomposition, which isn't exploiting the structure.

README.md
gpleiss: Fixed

README.md
gpleiss: fixed

README.md
gpleiss: Whoops - if this becomes `LowRankRootLinearOperator`, then it uses Woodbury. I'm going to fix that now.

README.md
gpleiss: included

README.md
gpleiss: This exists in the [Using LinearOperator objects](https://github.com/cornellius-gp/linear_operator/blob/readme/README.md#using-linearoperator-objects) section - but do you have any suggestions for where to put it more prominently?

README.md
gpleiss: Suggestions for where to put it?

README.md
Balandat: I think it's fine to just add a comment here in this list that states this (possibly referring the reader to the details below)

README.md
Balandat: you could add another list item above so it's not as split: 

1. specifies a `_transpose_nonbatch` function (the adjoint of the LinearOperator).
1. (optionally) defines other functions (e.g. `logdet`, `eigh`, etc.) to accelerate computations for which efficient structure-exploiting routines exist.

Balandat (approved):

PR #26, opened 2022-10-18
Update LO.__getitem__ to handle broadcasted tensor indices.

gpleiss:
In GPyTorch, we occasionally use broadcasted tensor indices such as the following:

- idx_a - size [18, 4, 1]
- idx_b - size [18, 1, 4]

covar_matrix[..., idx_a, idx_b]  # Returns a matrix of size [... x 18 x 4 x 4].

This PR ensures that we can use broadcasting tensor indices (especially those that introduce new dimensions).

[Needed for https://github.com/cornellius-gp/gpytorch/pull/2140]

### DIFF
--- a/linear_operator/operators/_linear_operator.py
+++ b/linear_operator/operators/_linear_operator.py
@@ -2634,8 +2634,7 @@ class LinearOperator(ABC):
             tensor_index_shape = torch.broadcast_shapes(*[idx.shape for idx in orig_indices if torch.is_tensor(idx)])
             # Flatten existing tensor indices
             flattened_orig_indices = [
-                idx.expand(tensor_index_shape).reshape(-1) if torch.is_tensor(idx) else idx
-                for idx in orig_indices
+                idx.expand(tensor_index_shape).reshape(-1) if torch.is_tensor(idx) else idx for idx in orig_indices
             ]
             # Convert all indices into tensor indices
             (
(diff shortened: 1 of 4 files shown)


Balandat:
Hmm would this help with the long-standing bug https://github.com/cornellius-gp/gpytorch/issues/1853 ?

linear_operator/operators/_linear_operator.py
```diff
@@ -2623,13 +2629,27 @@ def __getitem__(
         # Call self._getitem - now that the index has been processed
         # Alternatively, if we're using tensor indices and losing dimensions, use self._get_indices
         if row_col_are_absorbed:
+            # Get broadcasted size of existing tensor indices
+            orig_indices = [*batch_indices, row_index, col_index]
+            tensor_index_shape = torch.broadcast_shapes(*[idx.shape for idx in orig_indices if torch.is_tensor(idx)])
+            # Flatten existing tensor indices
+            flattened_orig_indices = [
+                idx.expand(tensor_index_shape).contiguous().view(-1) if torch.is_tensor(idx) else idx
```
Balandat: Use `reshape(-1)` here instead if we make a `contiguous` call anyway?

linear_operator/operators/cat_linear_operator.py
```diff
@@ -244,6 +244,14 @@ def _getitem(
                     res_list.append(res)
 
         elif torch.is_tensor(cat_dim_indices):
+            if cat_dim_indices.dim() > 1:
+                # Supporting broadcasting tensor indices for CatLinearOpeartor is very challenging to implement,
+                # and probvably not a featuer we need.
```
Balandat: provably not a future we need? 

```suggestion
                # and probably not a feature we need.
```

Balandat (approved):
gtmtl

linear_operator/operators/_linear_operator.py
gpleiss:
```suggestion
                idx.expand(tensor_index_shape).reshape(-1) if torch.is_tensor(idx) else idx
```

PR #27, opened 2022-10-19
Adds settings.trace_mode and skip cholesky jitter

feynmanliang:
We would like to [use gpytorch in our NUTS sampler](https://github.com/facebookresearch/beanmachine/blob/main/tests/ppl/experimental/gp/inference_test.py#L61-L70). While the linked code works, it does not utilize JIT compilation (`use_nnc`) because of control flow.

I understand `trace_mode` was originally used to disable custom pytorch functions back when JIT did not support compiling through them. This PR (along with https://github.com/cornellius-gp/gpytorch/pull/2163) revives the setting for the AoT JIT compiler (https://pytorch.org/functorch/nightly/aot_autograd.html).

 * Moves `settings.trace_mode` from `gpytorch` to `linear_operators`
 * Skips conditional control flow when `trace_mode` is true, currently:
   * PSD check / jitter in `linear_operators.utils.cholesky`
   * Diagonal `x1_eq_x2` check in `gpytorch.kernels.kernel`


gpleiss (approved):

gpleiss:
This makes sense to me! It seems like we'd want a PR in GPyTorch to ensure that `gpytorch.settings.trace_mode` also calls `linear_operator.settings.trace_mode`.

feynmanliang:
Sounds good, I will send a commit for the unit tests soon.

PR #31, opened 2022-11-01
Register torch.isclose

Balandat:
This is called in some places inside the torch.distributions code when `validate_args=True`.

This is a pretty basic implementation and we probably don't want to do this kind of things if the linear operators are very large since `islose` returns a tensor of the same size as the (often implicitly represented) operator.

**TODO:** Add unit tests

### DIFF
--- a/linear_operator/operators/_linear_operator.py
+++ b/linear_operator/operators/_linear_operator.py
@@ -2716,7 +2716,7 @@ class LinearOperator(ABC):
         # Regardless, if possible it would make sense to overwrite this method on the subclasses if that can
         # be done without instantiating the full tensor.
         warnings.warn(
-            f"Converting {self.class.name} into a dense torch.Tensor due to a torch.isclose call. "
+            f"Converting {self.__class__.__name__} into a dense torch.Tensor due to a torch.isclose call. "
             "This may incur substantial performance and memory penalties.",
             PerformanceWarning,
         )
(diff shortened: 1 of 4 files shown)


linear_operator/operators/_linear_operator.py
```diff
@@ -2706,6 +2710,13 @@ def __getitem__(
         # We're done!
         return res
 
+    def _isclose(self, other, rtol: float = 1e-05, atol: float = 1e-08, equal_nan: bool = False) -> Tensor:
+        # As the default we can fall back to just calling isclose on the dense tensors. This is problematic
+        # if the represented tensor is massive (in which case using this method may not make a lot of sense.
+        # Regardless, if possible it would make sense to overwrite this method on the subclasses if that can
+        # be done without instantiating the full tensor.
+        return torch.isclose(to_dense(self), to_dense(other), rtol=rtol, atol=atol, equal_nan=equal_nan)
```
gpleiss, line 2723: I wonder if we should throw a warning if we call this default method. Something like:

"Converting {self.__class__.__name__} into a dense torch.Tensor due to a `torch.isclose` call. It is likely that this is due to argument validation, potentially from a `torch.distributions.MultivariateNormal` object"

Thoughts?

gpleiss (approved):

linear_operator/operators/_linear_operator.py
Balandat: I think we can issue a warning here if we convert a (non-dense) tensor. I think pointing to MVN is probably way to specific, I'm sure `torch.isclose` is used in many other ways and mentioning MVN could just confuse a lot of people.

PR #32, opened 2022-11-02
Introduce reshape

Balandat:
This function is used ubiquitously with tensors, so in order to make `LinearOperator` a drop-in replacement for tensors we need to use it.

`reshape` (other than view) will reallocate new memory if necessary - this concept doesn't really make a to of sense for `LinearOperator`s, so a straightforward thing to do is just to make this an alias for `expand` and not allow reshaping the last two (non-batch) dimensions for now. The current implementation of course only makes sense for square `LinearOperator`s (or rather operators where the non-batch dimensions don't change).

### DIFF
--- a/linear_operator/operators/_linear_operator.py
+++ b/linear_operator/operators/_linear_operator.py
@@ -1972,8 +1972,10 @@ class LinearOperator(ABC):
         """
         Alias for expand
         """
-        # TODO: Do we really want to do this? I guess the difference betwen reshape and
-        # expand is kind of unclear for LinearOperator.
+        # While for regular tensors expand doesn't handle a leading non-existing -1 dimension,
+        # reshape does. So we handle this conversion here.
+        if len(sizes) == len(self.shape) + 1 and sizes[0] == -1:
+            sizes = (1,) + sizes[1:]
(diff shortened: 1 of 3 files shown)


linear_operator/operators/_linear_operator.py
```diff
@@ -1964,6 +1964,14 @@ def requires_grad_(self, val: bool) -> LinearOperator:
         self._set_requires_grad(val)
         return self
 
+    def reshape(self, *sizes: Union[torch.Size, Tuple[int, ...]]) -> LinearOperator:
+        """
+        Alias for expand
+        """
+        # TODO: Do we really want to do this? I guess the difference betwen reshape and
+        # expand is kind of unclear for LinearOperator.
+        return self.expand(*sizes)
```
Balandat, line 1979: TODO: Raise an informative warning if this tries to make modify the last two dimensions (that we have to think about more carefully when supporting non-square tensors more broadly).

gpleiss: Currently, "expand" errors out when you expand the last two dimensions, so I think this PR is okay as is without a warning.

gpleiss (approved):

Issue #33, opened 2022-11-03
[Bug] `mul` with mixed LinearOperator and Tensor gives incorrect shape


dannyfriar:
# 🐛 Bug

Calling the `LinearOperator` `mul` method on a `torch.Tensor` results in the wrong shape (although converting to a regular tensor appears to give the correct result). This caused GPyTorch to crash when using a `ScaleKernel` with an `AdditiveKernel` (in the multi-task case only).

## To reproduce

```python
import torch
from linear_operator import to_linear_operator

a = torch.randn(2, 1, 1)
b = torch.randn(1, 1000, 10000)

print(b.mul(a).shape)

a_lo = to_linear_operator(a)
b_lo = to_linear_operator(b)

print(b_lo.mul(a).shape)
print(b_lo.mul(a).to_dense().shape)
```

** Output **
```
torch.Size([2, 1000, 10000])
torch.Size([1, 1000, 10000])
torch.Size([2, 1000, 10000])
```

## System information
linear-operator==0.1.1

dannyfriar:
Attempted a fix here: https://github.com/cornellius-gp/linear_operator/pull/35. It seems (from the docstring of `ConstantMulLinearOperator`) that this should only be used when the batch dimensions are the same.

PR #35, opened 2022-11-05
Fix linear operator mul torch.Tensor

dannyfriar:
See https://github.com/cornellius-gp/linear_operator/issues/33


Balandat (commented):
Thanks for the fix, this looks good to me. Would you be able to add a unit test covering this behavior as well?

dannyfriar:
Sure, I've added a test for this case

Balandat:
Thanks - lgtm except for a lint issue.

Balandat (approved):

PR #39, opened 2022-11-07
Adding Permutation Linear Operators

SebastianAment:
This PR adds `PermutationLinearOperator` and `TransposePermutationLinearOperator`. The first one represents a generic 2d permutation matrix, while the second one is a special type of permutation `P` that satisfies
```python
P @ X.flatten() = vec(X.transpose(-2, -1)).
```

This transpose permutation shows up in recent Monarch matrices, as well as the structured representations of kernel matrices of gradient kernels.

A caveat is that the test suites do not inherit from `LinearOperatorTestCase`, since that leads to a lot of errors mainly due to missing support for changing batch shapes. Especially for the transpose permutation, adding support for `expand` and `reshape` seems unnatural, since it would require the introduction of a `batch_shape` field, that is completely independent of the functioning of the operator except to make it play nice with the existing testing infrastructure. 

Let me know what you think!

### DIFF
--- a/linear_operator/operators/permutation_linear_operator.py
+++ b/linear_operator/operators/permutation_linear_operator.py
@@ -1,0 +1,168 @@
+from typing import Optional, Tuple
+
+import torch
+from torch import Tensor
+
+from ._linear_operator import LinearOperator
+
+
+class AbstractPermutationLinearOperator(LinearOperator):
+    r"""Abstract base class for permutation operators.
+    Incorporates 1) square shape, 2) common input shape checking, and
+    3) the fact that permutation matrices' transposes are their inverses.
+    """
+
+    def inverse(self):
+        return self._transpose_nonbatch()
+
+    def _solve(self, rhs: Tensor):
+        self._matmul_check_shape(rhs)
+        return self.inverse() @ rhs
+
+    def _matmul_check_shape(self, rhs: Tensor) -> None:
+        if rhs.shape[-2] != self.shape[-1]:
+            raise ValueError(
+                f"{rhs.shape[0] = } incompatible with first dimensions of"
+                f"permutation operator with shape {self.shape}."
+            )
+
+    def _matmul_batch_shape(self, rhs: Tensor) -> torch.Size:
+        return torch.broadcast_shapes(self.batch_shape, rhs.shape[:-2])
+
+
+class PermutationLinearOperator(AbstractPermutationLinearOperator):
+    r"""LinearOperator that lazily represents a permutation matrix with O(n) memory.
+    Upon left-multiplication, it permutes the first non-batch dimension of a tensor.
+
+    Args:
+        - perm: A permutation tensor with which to permute the first non-batch
+            dimension through matmul. Should have integer elements. If perm
+            is multi-dimensional, the corresponding operator MVM broadcasts
+            the permutation along the batch dimensions. That is, if perm is two-
+            dimensional, perm[i, :] should be a permutation for every i.
+        - inv_perm: Optional tensor representing the inverse of perm, is computed
+            via a O(n log(n)) sort if not given.
+        - validate_args: Boolean
+    """
+
+    def __init__(
+        self,
+        perm: Tensor,
+        inv_perm: Optional[Tensor] = None,
+        validate_args: bool = True,
+    ):
+        if not isinstance(perm, Tensor):
+            raise ValueError("perm is not a Tensor.")
+
+        if inv_perm is not None:
+            if perm.shape != inv_perm.shape:
+                raise ValueError("inv_perm does not have the same shape as perm.")
+
+            batch_indices = self._batch_indexing_helper(perm.shape[:-1])
+            sorted_perm = perm[batch_indices + (inv_perm,)]
+        else:
+            sorted_perm, inv_perm = perm.sort(dim=-1)
+
+        if validate_args:
+            if torch.is_floating_point(sorted_perm) or torch.is_complex(sorted_perm):
+                raise ValueError("perm does not have integer elements.")
+
+            for i in range(sorted_perm.shape[-1]):
+                if (sorted_perm[..., i] != i).any():
+                    raise ValueError(
+                        f"Invalid perm-inv_perm input, index {i} missing or not at "
+                        f"correct index for permutation with {perm.shape = }."
+                    )
+
+        self.perm = perm
+        self.inv_perm = inv_perm
+        super().__init__(perm, inv_perm, validate_args=validate_args)
+
+    def _matmul(self, rhs: Tensor) -> Tensor:
+        # input rhs is guaranteed to be at least two-dimensional due to matmul implementation
+        self._matmul_check_shape(rhs)
+
+        # batch broadcasting logic
+        batch_shape = self._matmul_batch_shape(rhs)
+        expanded_rhs = rhs.expand(*batch_shape, *rhs.shape[-2:])
+        ndim = expanded_rhs.ndim
+
+        batch_indices = self._batch_indexing_helper(batch_shape)
+        batch_indices = tuple(index.unsqueeze(-1) for index in batch_indices)  # expanding to non-batch dimensions
+        perm_indices = self.perm.unsqueeze(-1)
+        final_indices = torch.arange(rhs.shape[-1]).view((1,) * (ndim - 1) + (-1,))
+        indices = batch_indices + (perm_indices, final_indices)
+        return expanded_rhs[indices]
+
+    def _batch_indexing_helper(self, batch_shape: torch.Size) -> Tuple:
+        """Creates a tuple of indices with broadcastable shapes to preserve the
+        batch dimensions when indexing into the non-batch dimensions with `perm`.
+
+        Args:
+            - batch_shape: the batch shape for which to generate the broadcastable indices.
+        """
+        return tuple(
+            torch.arange(n).view(
+                (1,) * i + (-1,) + (1,) * (len(batch_shape) - i - 1) + (1,)  # adding one non-batch dimension
+            )
+            for i, n in enumerate(batch_shape)
+        )
+
+    def _size(self) -> torch.Size:
+        return torch.Size((*self.perm.shape, self.perm.shape[-1]))
+
+    def _transpose_nonbatch(self):
+        return PermutationLinearOperator(perm=self.inv_perm, inv_perm=self.perm, validate_args=False)
+
+    def to_sparse(self) -> Tensor:
+        """Returns a sparse CSR tensor that represents the PermutationLinearOperator."""
+        # crow_indices[i] is index where values of row i begin
+        return torch.sparse_csr_tensor(
+            crow_indices=torch.arange(self.shape[-1] + 1).expand(*self.batch_shape, -1).contiguous(),
+            col_indices=self.perm,
+            values=torch.ones_like(self.perm),
+        )
+
+
+class TransposePermutationLinearOperator(AbstractPermutationLinearOperator):
+    r"""LinearOperator that represents a permutation matrix `P` with O(1) memory.
+    In particular, P satisfies
+
+        `P @ X.flatten(-2, -1) = X.transpose(-2, -1).flatten(-2, -1)`,
+
+    where `X` is an `m x m` matrix and P has size `n x n` where `n = m^2`.
+
+    Args:
+        - m: dimension on which the transpose operation is taking place. The size of
+            the permutation matrix that the operator represents is then `n = m^2`.
+    """
+
+    def __init__(self, m: int):
+        if m < 1:
+            raise ValueError(f"m = {m} has to be a positive integer.")
+        super().__init__(m=m)
+        self.n = m * m  # size of implicitly represented linear operator
+        self.m = m  # (m, m) is size of the reshaped input which is transposed
+        self._dtype = type(m)
+
+    def _matmul(self, rhs: Tensor) -> Tensor:
+        self._matmul_check_shape(rhs)
+        return rhs.unflatten(dim=-2, sizes=(self.m, self.m)).transpose(-3, -2).flatten(start_dim=-3, end_dim=-2)
+
+    def _size(self) -> torch.Size:
+        return torch.Size((self.n, self.n))
+
+    def _transpose_nonbatch(self):
+        return self
+
+    @property
+    def dtype(self):
+        return self._dtype
+
+    def type(self, dtype: torch.dtype) -> LinearOperator:
+        self._dtype = dtype
+        return self
+
+    @property
+    def device(self):
+        return None
--- a/test/operators/test_permutation_linear_operator.py
+++ b/test/operators/test_permutation_linear_operator.py
@@ -1,0 +1,94 @@
+#!/usr/bin/env python3
+
+import unittest
+
+import torch
+
+from linear_operator.operators import PermutationLinearOperator, TransposePermutationLinearOperator
+
+
+class TestPermutationLinearOperator(unittest.TestCase):
+    def test_permutation_linear_operator(self):
+        with self.assertRaisesRegex(ValueError, "perm is not a Tensor."):
+            PermutationLinearOperator([1, 3, 5])
+
+        with self.assertRaisesRegex(ValueError, "Invalid perm*"):
+            PermutationLinearOperator(torch.tensor([1, 3, 5]))
+
+        with self.assertRaisesRegex(ValueError, "Invalid perm*"):
+            PermutationLinearOperator(torch.tensor([0, 2, 1]), torch.tensor([0, 1, 2]))
+
+        n = 3
+        P = PermutationLinearOperator(torch.randperm(n))
+
+        b1 = 2
+        b2 = 5
+        batch_shapes = [(), (b1,), (b2, b1)]
+
+        permutations = [
+            torch.randperm(n),
+            torch.cat(tuple(torch.randperm(n).unsqueeze(0) for _ in range(b1)), dim=0),
+        ]
+        operators = [PermutationLinearOperator(perm) for perm in permutations]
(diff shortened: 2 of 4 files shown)


linear_operator/operators/permutation_linear_operator.py
```diff
@@ -0,0 +1,133 @@
...
+    r"""LinearOperator that represents a permutation matrix `P` with O(1) memory.
+    In particular, P satisfies
+
+        `P @ X.flatten() = vec(X.transpose(-2, -1))`,
+
+    where `X` is an `m x m` matrix and P has size `n x n` where `n = m^2`.
+
+    Args:
+        - n: size of the permutation matrix that the operator represents.
+            NOTE: Has to be a perfect square, i.e. there is an `m` such that `n = m^2`.
```
SebastianAment: ... there is an *integer* `m` such that ...

linear_operator/operators/permutation_linear_operator.py
```diff
@@ -0,0 +1,133 @@
...
+        return self.inverse() @ rhs
+
+    def _size(self):
+        n = len(self)
+        return (n, n)
+
+    def _check_shape(self, rhs) -> None:
+        if rhs.shape[-2] != len(self):
+            raise ValueError(
+                f"rhs.shape[-2] = {rhs.shape[-2]} incompatible " f"with size of permutation operator {len(self)}."
```
Balandat:
```suggestion
                f"rhs.shape[-2] = {rhs.shape[-2]} incompatible "
                f"with size of permutation operator {len(self)}."
```

linear_operator/operators/permutation_linear_operator.py
```diff
@@ -0,0 +1,133 @@
...
+        self,
+        perm: Tensor,
+        inv_perm: Optional[Tensor] = None,
+        check: bool = True,
+    ):
+        if not isinstance(perm, Tensor):
+            raise ValueError("perm is not a Tensor.")
+
+        if not perm.ndim == 1:
+            raise ValueError("perm is not a 1d Tensor.")
```
Balandat: Would we at some point want to support a batched permutation tensor? Say where `perm` is of size `(batch_shape) x n` where `perm[i, :]` indicates the permutation of the `i`-th batch (here `i` could be multidimensional as well).

linear_operator/operators/permutation_linear_operator.py
```diff
@@ -0,0 +1,133 @@
...
+        - inv_perm: Optional tensor representing the inverse of perm, is computed
+            via a O(n log(n)) sort if not given.
+        - check: Boolean
+    """
+
+    def __init__(
+        self,
+        perm: Tensor,
+        inv_perm: Optional[Tensor] = None,
+        check: bool = True,
```
Balandat: how about calling this `validate_args` to be consistent with the torch.distributions naming?

linear_operator/operators/permutation_linear_operator.py
```diff
@@ -0,0 +1,133 @@
...
+        if inv_perm is not None:
+            sorted_perm = perm[inv_perm]
+        else:
+            sorted_perm, inv_perm = perm.sort()
+
+        if check:
+            for i, si in enumerate(sorted_perm):
+                if torch.is_floating_point(si) or torch.is_complex(si):
+                    raise ValueError("perm does not have integer elements.")
+                if sorted_perm[i] != i:
```
Balandat: no need to do this if `inv_perm` is `None`.

linear_operator/operators/permutation_linear_operator.py
```diff
@@ -0,0 +1,133 @@
...
+            raise ValueError("perm is not a 1d Tensor.")
+
+        if inv_perm is not None:
+            sorted_perm = perm[inv_perm]
+        else:
+            sorted_perm, inv_perm = perm.sort()
+
+        if check:
+            for i, si in enumerate(sorted_perm):
+                if torch.is_floating_point(si) or torch.is_complex(si):
```
Balandat: You can call these directly on `perm` and `inv_perm` instead of on the individual elements

linear_operator/operators/permutation_linear_operator.py
```diff
@@ -0,0 +1,133 @@
...
+                if sorted_perm[i] != i:
+                    raise ValueError(
+                        f"Invalid perm-inv_perm input, index {i} missing or not at "
+                        f"correct index for permutation of length {len(perm)}."
+                    )
+        self.perm = perm
+        self.inv_perm = inv_perm
+        super().__init__(perm, inv_perm, check=check)
+
+    def _matmul(self, rhs: Tensor):
```
Balandat:
```suggestion
    def _matmul(self, rhs: Tensor) -> Tensor:
```

linear_operator/operators/permutation_linear_operator.py
```diff
@@ -0,0 +1,133 @@
...
+        self._check_shape(rhs)
+        return rhs[..., self.perm, :]
+
+    def __len__(self):
```
Balandat:
```suggestion
    def __len__(self) -> int:
```

linear_operator/operators/permutation_linear_operator.py
```diff
@@ -0,0 +1,133 @@
...
+        return len(self.perm)
```
Balandat: Is defining this necessary here? Many `LinearOperator`s don't have a `__len__` defined, and the one here is inconsistent with the usual `__len__` behavior of dense tensors (where this would give the size of the leading dimension).

Balandat: Seems like you should be able to just call `len(self.perm)` internally as needed in a few places

Balandat: You could just use `.shape[-1]` here

linear_operator/operators/permutation_linear_operator.py
```diff
@@ -0,0 +1,133 @@
...
+        for (i, pi) in self.perm:
+            P[pi, i] = 1
+        return P
+
+
+class TransposePermutationLinearOperator(AbstractPermutationLinearOperator):
+    r"""LinearOperator that represents a permutation matrix `P` with O(1) memory.
+    In particular, P satisfies
+
+        `P @ X.flatten() = vec(X.transpose(-2, -1))`,
```
Balandat: If `X` were batched then `flatten()` would not work here so if you want batch support you could do this here
```suggestion
        `P @ X.flatten(-2, -1) = vec(X.transpose(-2, -1))`,
```

linear_operator/operators/permutation_linear_operator.py
```diff
@@ -0,0 +1,133 @@
...
+
+    where `X` is an `m x m` matrix and P has size `n x n` where `n = m^2`.
+
+    Args:
+        - n: size of the permutation matrix that the operator represents. NOTE: `n` has
+            to be a perfect square, i.e. there is an integer `m` such that `n = m^2`.
```
Balandat: why not make `m` the arg to the constructor?

linear_operator/operators/permutation_linear_operator.py
```diff
@@ -0,0 +1,133 @@
...
+    def __init__(self, n: int):
+        m = isqrt(n)  # integer square root
+        if not m * m == n:
+            raise ValueError(f"n = {n} is not a perfect square")
+
+        self.n = n  # size of implicitly represented linear operator
+        self.m = m  # (m, m) is size of the reshaped input which is transposed
+        super().__init__(n=self.n)
+
+    def _matmul(self, rhs: Tensor):
```
Balandat:
```suggestion
    def _matmul(self, rhs: Tensor) -> Tensor:
```

linear_operator/operators/permutation_linear_operator.py
Balandat: Should this not also work when `rhs` is a `LinearOperator`?

linear_operator/operators/permutation_linear_operator.py
```diff
@@ -0,0 +1,133 @@
...
+        return PermutationLinearOperator(perm=self.inv_perm, inv_perm=self.perm, check=False)
+
+    def inverse(self):
+        return self._transpose_nonbatch()
+
+    def to_dense(self):
+        P = torch.zeros(self.shape, dtype=self.perm.dtype)
+        for (i, pi) in self.perm:
+            P[pi, i] = 1
+        return P
```
Balandat:
```suggestion
    def to_dense(self) -> Tensor:
        P = torch.zeros(self.shape, dtype=self.perm.dtype)
        P[..., self.perm, torch.arange(len(self.perm))] = 1
        return P
```

Or, if you wanted to be self-referential,
```
return self @ torch.diag_embed(torch.ones(self.shape[:-1]))
```

linear_operator/operators/permutation_linear_operator.py
```diff
@@ -0,0 +1,133 @@
...
+
+    def _check_shape(self, rhs) -> None:
+        if rhs.shape[-2] != len(self):
+            raise ValueError(
+                f"rhs.shape[-2] = {rhs.shape[-2]} incompatible " f"with size of permutation operator {len(self)}."
+            )
+
+
+class PermutationLinearOperator(AbstractPermutationLinearOperator):
+    r"""LinearOperator that lazily represents a permutation matrix with O(n) memory.
```
Balandat, line 34: SHould be more clear that this permutes just the first non-batch dim here

linear_operator/operators/permutation_linear_operator.py
```diff
@@ -0,0 +1,133 @@
...
+    def _transpose_nonbatch(self):
+        return self
+
+    @property
+    def dtype(self):
+        return torch.int
+
+    @property
+    def device(self):
+        return torch.device
```
Balandat: which `device`? Is there a a "universal device" in torch for objects that don't actually have any (tensor) memory allocated? May also just return `None` here instead (which corresponds to the default device in terms of behavior)

test/operators/test_permutation_linear_operator.py
```diff
@@ -0,0 +1,70 @@
...
+        n = 5
+        P = PermutationLinearOperator(torch.randperm(n))
+        for x in [
+            torch.randn(n),
+            torch.randn(n, 3),
+            torch.randn(2, n, 3),
+            torch.randn(7, 2, n, 3),
+        ]:
+            # computed inverse permutation field sorts the permutation
+            self.assertTrue(torch.equal(P.perm[P.inv_perm], torch.tensor(range(n))))
```
Balandat:
```suggestion
            self.assertTrue(torch.equal(P.perm[P.inv_perm], torch.arange(n)))
```

linear_operator/operators/permutation_linear_operator.py
SebastianAment: I added "Upon left-multiplication, it permutes the first non-batch dimension of a tensor." in an upcoming commit.

linear_operator/operators/permutation_linear_operator.py
SebastianAment: If `inv_perm` is `None`, `sorted_perm` is guaranteed to be ordered correctly, but this can still fail if `perm` is not actually a permutation. I am testing this case with 
```python
with self.assertRaisesRegex(ValueError, "Invalid perm*"):
     PermutationLinearOperator(torch.tensor([1, 3, 5]))
```
in the test file. So I think there is still value in keeping this test. What do you think?

linear_operator/operators/permutation_linear_operator.py
SebastianAment: As far as I can tell, the `matmul` with a `LinearOperator` should always be represented lazily, which is the default behavior of `matmul` in `_linear_operator`. That means we'll only ever get `Tensor` types be passed into `_matmul`. As an example, Monarch matrices require the lazy `MatmulLinearOperator` representation of a `PermutationLinearOperator` and a `BlockDiagLinearOperator` to be computationally efficient. Is there another case I am forgetting about?

linear_operator/operators/permutation_linear_operator.py
SebastianAment: That's clean!

linear_operator/operators/permutation_linear_operator.py
SebastianAment: Done in the upcoming commit.

SebastianAment:
@Balandat The latest commit supports batched `PermutationLinearOperators` that you proposed [here](https://github.com/cornellius-gp/linear_operator/pull/39#discussion_r1020427301). The indexing logic required to make this happen increased the complexity of the implementation.

For `TransposePermutationOperator`, I also introduced a `batch_shape` argument, but am not sure if we should keep it. It is a step towards unifying its interface with the rest of the `LinearOperators`, but is a primarily syntactic addition.

I replaced the custom `to_dense` method with a `to_sparse` method, since the `to_dense` is already covered by the base class implementation. However, the PyTorch's CSR tensor implementation only allows batch support for `torch.__version__ > 1.12`, so I am toggling the corresponding test based on the torch version. The test passes locally.

linear_operator/operators/permutation_linear_operator.py
```diff
@@ -0,0 +1,168 @@
...
+    def _matmul_batch_shape(self, rhs: Tensor) -> torch.Size:
+        return torch.broadcast_shapes(self.batch_shape, rhs.shape[:-2])
+
+
+class PermutationLinearOperator(AbstractPermutationLinearOperator):
+    r"""LinearOperator that lazily represents a permutation matrix with O(n) memory.
+    Upon left-multiplication, it permutes the first non-batch dimension of a tensor.
+
+    Args:
+        - perm: A permutation tensor with which to permute the first non-batch
```
SebastianAment, line 38: I should add information regarding batched `perm` tensors here.

linear_operator/operators/permutation_linear_operator.py
```diff
@@ -0,0 +1,168 @@
...
+class TransposePermutationLinearOperator(AbstractPermutationLinearOperator):
+    r"""LinearOperator that represents a permutation matrix `P` with O(1) memory.
+    In particular, P satisfies
+
+        `P @ X.flatten(-2, -1) = X.transpose(-2, -1).flatten(-2, -1)`,
+
+    where `X` is an `m x m` matrix and P has size `n x n` where `n = m^2`.
+
+    Args:
+        - m: dimension on which the transpose operation is taking place. The size of
```
SebastianAment, line 136: If we decide to keep `batch_shape`, I should add it to the doc string here.

linear_operator/operators/permutation_linear_operator.py
```diff
@@ -0,0 +1,168 @@
...
+        return tuple(indices)
+
+    def _size(self) -> torch.Size:
+        return torch.Size((*self.perm.shape, self.perm.shape[-1]))
+
+    def _transpose_nonbatch(self):
+        return PermutationLinearOperator(perm=self.inv_perm, inv_perm=self.perm, validate_args=False)
+
+    def to_sparse(self):
+        # crow_indices[i] is index where values of row i begin
```
SebastianAment, line 119: Missing doc-string at this point.

linear_operator/operators/permutation_linear_operator.py
```diff
@@ -0,0 +1,168 @@
...
+
+        # batch broadcasting logic
+        batch_shape = self._matmul_batch_shape(rhs)
+        expanded_rhs = rhs.expand(*batch_shape, *rhs.shape[-2:])
+        ndim = expanded_rhs.ndim
+
+        batch_indices = self._batch_indexing_helper(batch_shape)
+        batch_indices = tuple(index[..., None] for index in batch_indices)  # expanding to non-batch dimensions
+        perm_indices = self.perm[..., None]
+        final_indices = torch.arange(rhs.shape[-1])[(None,) * (ndim - 1) + (slice(None),)]
```
Balandat: Not sure if cleaner but you could also do this
```suggestion
        final_indices = torch.arange(rhs.shape[-1]).view(*(1,)*(ndim-1), -1)
```

linear_operator/operators/permutation_linear_operator.py
```diff
@@ -0,0 +1,168 @@
...
+        for i, n in enumerate(batch_shape):
+            indices.append(
+                torch.arange(n)[
+                    (None,) * i
+                    + (slice(None),)
+                    + (None,) * (len(batch_shape) - i - 1)
+                    + (None,)  # adding one non-batch dimension
+                ]
+            )
+        return tuple(indices)
```
Balandat: Can make a tuple right away. Also could use `view` here instead (not necessary though)
```suggestion
        return tuple(
            torch.arange(n)[
                (None,) * i
                + (slice(None),)
                + (None,) * (len(batch_shape) - i - 1)
                + (None,)  # adding one non-batch dimension
            ]
            for i, n in enumerate(batch_shape)
        )
```

linear_operator/operators/permutation_linear_operator.py
Balandat: Hmm the only thing batch shape does is to give it a specific `size`? I don't think that's necessary so long as the operator is properly broadcastable (which it seems it is). Not sure if we want to ever represent the permutation with a specific batch shape. @gpleiss any opinions here?

SebastianAment:
I changed the `None`-based indexing to `view`-based indexing, which I think improves readability. Thanks for the suggestion @Balandat ! 

I also removed `batch_shape` from `TransposePermutationLinearOperator`. Its only value is to unify its interface with other `LinearOperators` and allow e.g. `expand` calls. I'll put it back in if the need ever arises, but think we can ship this version first.

If you approve the current version, could you hold off on merging directly? I would like to squash the commits before.

Balandat:
> If you approve the current version, could you hold off on merging directly? I would like to squash the commits before.

Sure, please go ahead. Good to keep things tidy :)

Balandat (approved):

SebastianAment:
Good to go! :)

SebastianAment:
One unexpectedly adventurous rebase later, this really is now good to go.

PR #41, opened 2022-11-08
Consolidating `DiagLinearOperator`'s `matmul`

SebastianAment:
This commit moves the standard case of multiplying a `DiagLinearOperator` with a `Tensor` to the public `matmul` method from the private `_matmul`, thereby consolidating all implementation details in one method. 

A few notes on the change:

- putting the `Tensor` case at the top of the `matmul` method means there will be minimal overhead through indirection.
- `matmul` needs to be public to treat the special cases before the generic `matmul` creates a `MatmulLinearOperator`. 
- we can't remove the private `_matmul` because operators that do not have special cases in the `matmul` of `DiagLinearOperator` (e.g. some of the kronecker products) still use the generic `matmul` code path, which ends up calling `_matmul`. 
- Instead of having a separate implementation, `_matmul` now calls `matmul`, thereby ensuring that all special cases will be considered whenever a `matmul` is carried out with a `DiagLinearOperator`.


Balandat (approved):
This lgtm, thanks!

Issue #3, opened 2022-08-09
`DenseLinearOperator` does not have `operator` attribute


SebastianAment:
I ran into this while porting over [this PR](https://github.com/cornellius-gp/gpytorch/pull/2093) here.

When I execute the test suite, I get 122 errors of the form 
```
AttributeError: 'DenseLinearOperator' object has no attribute 'operator'
```
When I add an `operator` attribute [here](https://github.com/cornellius-gp/linear_operator/blob/ea4f9c85e92ad3f76c87d38f272386f2dc202144/linear_operator/operators/dense_linear_operator.py#L27) and set it to `tsr`, all tests pass. 

It seems like @gpleiss is about to [merge this](https://github.com/cornellius-gp/gpytorch/pull/2027), and I am not sure if there are any other pending changes coming into this repository. Let me know if I should hold off until the merge is complete, or if I should make a PR for this.

SebastianAment:
My mistake, I inadvertently changed the test!

PR #42, opened 2022-12-16
Add stricter type checking and expected matrix sizes

corwinjoy:
As per the discussion: https://github.com/cornellius-gp/gpytorch/discussions/2180
I have upgraded the library to use Jaxtyping.
This allows a couple things:
1. The signatures can now include explicit matrix sizes, showing what kinds of results are expected / where broadcasting may be happening.
2. I have upgraded the operator tests to annotate each tested library with typeguard. So, only during testing, the type hints are tested for correctness to see if the functions actually behave the way the type hints say they will. This helps make the type hints much more accurate so I think its a good step forward.

Issues:
1. The jaxtyping library requires jax.  It should be not to hard to get rid of this dependency (it is on the maintainer's todo list) but that has not been done yet.
2. It would be nice to be able to add the storage type for the operator (e.g. dense or not) but this again requires upgrading jaxtyping.
3. I would also like to upgrade the jaxtyping math so I could have expressions like min(M, N) where M,N are array dimensions for some operators.
4. Because typeguard does not inherit type hint signatures, I've had to copy/paste these from the base _linear_operators class to the derived operators. I think it probably makes sense to create a tool to do this automatically to help with signature maintenance.


.pre-commit-config.yaml
```diff
@@ -29,7 +29,7 @@ repos:
     -   id: isort
         language_version: python3
         exclude: ^(build/*)|(docs/*)|(examples/*)
-        args: [-w120, -m3, --tc, --project=gpytorch]
+        args: [-w120, -m3, --tc, --project=gpytorch, --profile=black]
```
corwinjoy: BTW. You will want this regardless. Sometimes black & isort can get in an infinite loop fight where they cancel each other's changes. This flag tells isort to use black conventions.

Balandat (commented):
Thanks for this! How hard will it be to get rid of the `jax` dependency here? It's rather heavyweight and may not be available in all environments that run `linear_operator` (in fact, this is the case for us). So in order to merge this in I think we need to get rid of it.

.pre-commit-config.yaml
Balandat: An alternative to this would be to use µfmt: https://ufmt.omnilib.dev/en/stable/ - we've had very good experiences using this. Ofc this is orthogonal to the changes here.

corwinjoy:
@Balandat  I was going through the code and exploring making jax optional when I found the project maintainer has already done this!  So, I am very happy to report that the latest versions of jaxtyping no longer require jax.  See here: https://github.com/google/jaxtyping/pull/50

Balandat:
Excellent news!

corwinjoy:
Probably the online unit tests are failing because I didn't add jaxtyping to the requirements in setup.py.
I have gone ahead and updated setup.py to add jaxtyping to the requirements. (The latest version which does not require jax).

corwinjoy:
For item #3 on the wishlist:
"I would also like to upgrade the jaxtyping math so I could have expressions like min(M, N) where M,N are array dimensions for some operators."
See the issue I have opened with jaxtyping. I'm hoping the author may give some hints.
https://github.com/google/jaxtyping/issues/51

corwinjoy:
Continuing to improve this, I have added a routine in the main directory, "propagate_type_hints.py".  The idea here is to maintain the type hints in just a single place, LinearOperator. Then this routine can be used to push the signatures to the derived classes & unit tests will check that the types in the derived classes match the official / abstract signature. I think this will make this much more maintainable. This also required a couple tweaks in a few places where functions like _mul_constant used inconsistent argument names.

corwinjoy:
To explain, here is why I think it makes sense to propagate signatures from the base LinearOperator class to derived classes:
1. Consistency. Make sure the derived implementations are following the promised interface.
2. Visibility. Make it easy to see what the expected signature is, along with dimensions. Repeating the signature in the derived classes enhances readability.
3. Necessity. The way that jaxtyping and typeguard are written, they won't run type checks unless type annotations are present in the derived method signature. So, if we want to type check these methods we either have to copy over the signatures or find a way to extend these libraries. (The extension looks difficult, from what I could tell).

Balandat:
@esantorella, @j-wilson this may be of interest to you.

linear_operator/operators/added_diag_linear_operator.py
```diff
@@ -173,20 +185,26 @@ def _init_cache_for_non_constant_diag(self, eye: Tensor, batch_shape: torch.Size
...
+        return_evals_as_lazy: Optional[bool] = False,
+    ) -> Tuple[Float[Tensor, "*batch M"], Optional[Float[LinearOperator, "*batch N M"]]]:
         if isinstance(self._diag_tensor, ConstantDiagLinearOperator):
             evals_, evecs = self._linear_op._symeig(eigenvectors=eigenvectors)
             evals = evals_ + self._diag_tensor._diagonal()
             return evals, evecs
         return super()._symeig(eigenvectors=eigenvectors)
 
-    def evaluate_kernel(self) -> LinearOperator:
+    def evaluate_kernel(self):
```
dannyfriar, line 208: Does this need a return type annotation?

linear_operator/operators/_linear_operator.py
```diff
@@ -728,7 +747,12 @@ def _set_requires_grad(self, val: bool) -> None:
                 if arg.dtype in (torch.float, torch.double, torch.half):
                     arg.requires_grad_(val)
 
-    def _solve(self, rhs: torch.Tensor, preconditioner: Callable, num_tridiag: int = 0) -> torch.Tensor:
+    def _solve(
+        self: Float[LinearOperator, "... N N"],
+        rhs: Float[torch.Tensor, "... N C"],
+        preconditioner: Optional[Callable] = None,
```
dannyfriar: Maybe complete the `Callable` annotation so it's clearer what the signature should look like

linear_operator/operators/block_diag_linear_operator.py
```diff
@@ -1,21 +1,22 @@
...
+from ._linear_operator import IndexType, LinearOperator
 from .block_linear_operator import BlockLinearOperator
 
 
 # metaclass of BlockDiagLinearOperator, overwrites behavior of constructor call
 # _MetaBlockDiagLinearOperator(base_linear_op, block_dim=-3) to return a DiagLinearOperator
 # if base_linear_op is a DiagLinearOperator itself
 class _MetaBlockDiagLinearOperator(ABCMeta):
-    def __call__(cls, base_linear_op: LinearOperator, block_dim=-3):
+    def __call__(cls, base_linear_op: Union[LinearOperator, Tensor], block_dim=-3):
```
dannyfriar:
```suggestion
    def __call__(cls, base_linear_op: Union[LinearOperator, Tensor], block_dim: int = -3):
```

linear_operator/operators/block_diag_linear_operator.py
```diff
@@ -56,7 +57,7 @@ def __init__(self, base_linear_op, block_dim=-3):
             )
 
     @property
-    def num_blocks(self):
+    def num_blocks(self) -> int:
         return self.base_linear_op.size(-3)
 
     def _add_batch_dim(self, other):
```
dannyfriar:
```suggestion
    def _add_batch_dim(self, other: Float[torch.Tensor, "... N C"]) -> Float[torch.Tensor, "... C"]:
```

dannyfriar: Not sure if this can be expressed properly with `jaxtyping`?

linear_operator/operators/block_diag_linear_operator.py
```diff
@@ -68,23 +69,25 @@ def _add_batch_dim(self, other):
...
+    def _cholesky(
+        self: Float[LinearOperator, "*batch N N"], upper: Optional[bool] = False
+    ) -> Float[LinearOperator, "*batch N N"]:
         from .triangular_linear_operator import TriangularLinearOperator
 
         chol = self.__class__(self.base_linear_op.cholesky(upper=upper))
         return TriangularLinearOperator(chol, upper=upper)
 
-    def _cholesky_solve(self, rhs, upper: bool = False):
+    def _cholesky_solve(self, rhs, upper: Optional[bool] = False) -> Union[LinearOperator, Tensor]:
```
dannyfriar: Type annotation for `rhs`

linear_operator/test/linear_operator_test_case.py
```diff
@@ -606,7 +615,11 @@ def test_add_jitter(self):
         linear_op = self.create_linear_op()
         evaluated = self.evaluate_linear_op(linear_op)
 
-        res = linear_operator.add_jitter(linear_op, 0.4).to_dense()
+        try:
```
dannyfriar: Why are these try/excepts needed?

propagate_type_hints.py
```diff
@@ -0,0 +1,111 @@
+# Propagate type hints & signatures defined in _linear_operator.py to derived classes.
```
dannyfriar: Maybe add the usage of this script to the `CONTRIBUTING.md` guide along with any other notes on typing

propagate_type_hints.py
```diff
@@ -0,0 +1,111 @@
...
+# This way we can enforce consistency between the base class signature and derived signatures.
+
+from pathlib import Path
+from typing import Dict, List, Optional, Tuple
+
+import libcst as cst
+
+
+class TypingCollector(cst.CSTVisitor):
+    def __init__(self):
```
dannyfriar:
```suggestion
    def __init__(self) -> None:
```

propagate_type_hints.py
```diff
@@ -0,0 +1,111 @@
...
+    def leave_FunctionDef(self, node: cst.FunctionDef) -> None:
+        self.stack.pop()
+
+
+class TypingTransformer(cst.CSTTransformer):
+
+    # List of LinearOperator functions we do not want to propagate the signature from
+    excluded_functions = ["__init__", "_check_args", "__torch_function__"]
+
+    def __init__(self, annotations):
```
dannyfriar: Nit: maybe define an `Annotations` type that can be used for this arg

dannyfriar (commented):
The signatures make things a lot clearer IMO. I've added some small comments mostly around a few missing annotations

linear_operator/operators/added_diag_linear_operator.py
corwinjoy: Probably not. Will remove.

linear_operator/operators/block_diag_linear_operator.py
corwinjoy: Maybe it can. Talking with the jaxtyping author, functions and operators are supported over dimensions. The function definition is reproduced below:
```
    def _add_batch_dim(self, other):
        *batch_shape, num_rows, num_cols = other.shape
        batch_shape = list(batch_shape)

        batch_shape.append(self.num_blocks)
        other = other.view(*batch_shape, num_rows // self.num_blocks, num_cols)
        return other
```

Which suggests the following signature using the div operator. (This passes runtime tests):

```
def _add_batch_dim(self: Float[LinearOperator, "*batch1 M P"], other: Float[torch.Tensor, "*batch2 N C"]) \
            -> Float[torch.Tensor, "batch2 N//batch1 C"]:
```

@dannyfriar What do you think?

linear_operator/operators/block_diag_linear_operator.py
corwinjoy: Looks good.

corwinjoy: OK. I have changed this to a more detailed definition:
```
def _cholesky_solve(self: Float[LinearOperator, "*batch N N"],
                        rhs: Float[LinearOperator, "batch N M"],
                        upper: Optional[bool] = False) \
            -> Union[Float[LinearOperator, "batch N M"], Float[Tensor, "batch N M"]]:
```

linear_operator/test/linear_operator_test_case.py
corwinjoy: Oops. Removed. This was some debugging I added to track down a function signature error.

propagate_type_hints.py
corwinjoy: Good idea. Next commit will contain updated notes in `CONTRIBUTING.md`

propagate_type_hints.py
corwinjoy: OK. I have upgraded this to use an Annotations type as a TypedDict per [PEP 589](https://peps.python.org/pep-0589/)

linear_operator/operators/_linear_operator.py
corwinjoy: @dannyfriar   OK. I have updated the Callable signature to be a function that takes a tensor in the shape of `rhs` and returns the same shape tensor. From what I can tell, the preconditional operates on the residual of the rhs. It passes tests but I'm not completely sure that is right. Please take a look.

```
def _solve(
        self: Float[LinearOperator, "... N N"],
        rhs: Float[torch.Tensor, "... N C"],
        preconditioner: Optional[Callable[[Float[torch.Tensor, "... N C"]], Float[torch.Tensor, "... N C"]]] = None,
        num_tridiag: Optional[int] = 0,
    ) -> Union[Float[torch.Tensor, "... N C"], Tuple[Float[torch.Tensor, "... N C"], Float[torch.Tensor, "... N N"]]]:

```

corwinjoy:
@esantorella, @gpleiss 
I think this all makes sense and looks good. Any thoughts?

linear_operator/operators/block_diag_linear_operator.py
dannyfriar: Sorry for the late reply I missed this. It looks good to me, clearer than what was there before

propagate_type_hints.py
```diff
@@ -0,0 +1,110 @@
...
+    base_visitor = collect_base_type_hints(base_filename)
+
+    pathlist = Path(directory).glob("*.py")
+    for path in pathlist:
+        if path.name[0] == "_":
+            continue
+        target = path
+        # target_out = Path(str(path) + "2")
+        target_out = target
+        print(f"Processing {(target.name, target_out.name)}")
```
Balandat:
```suggestion
        print(f"Processing {(target.name, target_out.name)}"). # noqa: T001
```

linear_operator/operators/_linear_operator.py
```diff
@@ -6,11 +6,14 @@
 import math
 import numbers
 import warnings
-from abc import ABC, abstractmethod
+from abc import abstractmethod
 from copy import deepcopy
+from types import EllipsisType
```
Balandat: This is python 3.10+ only. We'll need to support py3.8 for some time so this will require a different solution

corwinjoy: No problem. I have changed this to: `type("...")`

corwinjoy:
@esantorella, @gpleiss
Any thoughts on this PR?

gpleiss:
Sorry for the slow reply @corwinjoy - I’ll take a look at this on Monday!

linear_operator/operators/added_diag_linear_operator.py
```diff
@@ -67,21 +70,30 @@ def __init__(
         self._q_cache = None
         self._r_cache = None
 
-    def _matmul(self, rhs: Tensor) -> Tensor:
+    def _matmul(
+        self: Float[LinearOperator, "*batch M N"],
+        rhs: Union[Float[torch.Tensor, "*batch2 N C"], Float[torch.Tensor, "*batch2 N"]],
```
gpleiss, line 75: Why do we use `*batch` in certain settings versus `...` in others?

setup.py
```diff
@@ -37,7 +37,7 @@
     pass
 
 # Other requirements
-install_requires += ["scipy"]
+install_requires += ["scipy", "jaxtyping>=0.2.9"]
```
gpleiss: Can this be a dependency for testing, not for the main library?

test/operators/test_added_diag_linear_operator.py
```diff
@@ -3,13 +3,17 @@
 import unittest
 
 import torch
+from jaxtyping import install_import_hook
+
+with install_import_hook("linear_operator", ("typeguard", "typechecked")):
```
gpleiss: I am not familiar with jaxtyping. Does this context manager ensure that all methods in these files get type checked?

typing_examples.txt
```diff
@@ -0,0 +1,55 @@
+    # Predictable dimensions.
```
gpleiss: Is this file necessary? If so, could it live somewhere other than the base of the package repo?

typing_examples_jax.txt
```diff
@@ -0,0 +1,56 @@
+    # Predictable dimensions.
```
gpleiss: Same question here.

CONTRIBUTING.md
```diff
@@ -40,10 +40,24 @@ We use [standard sphinx docstrings](https://sphinx-rtd-tutorial.readthedocs.io/e
...
-While we currently do not enforce full consistency of these in our continuous integration
-test, you should strive to type check your code locally.
-For this we recommend using [pyre](https://pyre-check.org/).
-
+We are using [jaxtyping](https://github.com/google/jaxtyping) to help us be declarative about the dimension sizes used
+in the LinearOperator methods.
+The use of [jaxtyping](https://github.com/google/jaxtyping)  makes it clearer what the functions are doing algebraically
+and where broadcasting is happening.
+
+These type hints are checked in the unit tests by using
```
gpleiss, line 48: I noticed you didn't change our CI scripts at all. If I understand things correctly, the type hints are checked by the `install_import_hook` context manager?

linear_operator/test/linear_operator_test_case.py
```diff
@@ -617,7 +625,11 @@ def test_add_low_rank(self):
         new_rows = torch.randn(*linear_op.shape[:-1], 3)
 
         summed_lt = evaluated + new_rows.matmul(new_rows.mT)
-        new_lt = linear_op.add_low_rank(new_rows)
+        try:
+            new_lt = linear_op.add_low_rank(new_rows)
+        except Exception:
+            msg = traceback.format_exc()
```
gpleiss: What is the purpose of this?

linear_operator/test/type_checking_test_case.py
```diff
@@ -0,0 +1,97 @@
...
+    vec: Union[Float[torch.Tensor, "*batch N C"], Float[torch.Tensor, "*batch N"]],
+) -> Float[torch.Tensor, "*batch M C"]:
+    r"""
+    As above, but with bad return size
+    """
+    res = lo.matmul(vec)
+    return res
+
+
+class TestTypeChecking(unittest.TestCase):
```
gpleiss, line 73: This is awesome!

propagate_type_hints.py
gpleiss, line 1: Small nit: I'm not sure this file should live in the primary directory. Not sure where else to move it though...

gpleiss (commented):
This is an amazing effort @corwinjoy ! Sorry for the delay in reviewing it.

linear_operator/operators/added_diag_linear_operator.py
corwinjoy: A good question. This represents a current limitation in Jaxtyping. The idea here is that in _matmul *either* self or rhs could have a batch dimension. Then, the output dimension would have one of these batch sizes (if a batch was specified). So, I was hoping for a syntax like this:
```
def _matmul(
        self: Float[LinearOperator, "*batchA M N"],
        rhs: Float[torch.Tensor, "*batchB N C"],
    ) -> Float[torch.Tensor, "?batchA ?batchB M C"]:
```
In practice, this feature doesn't exist yet. It seems like there may be a way to get syntax like this into Jaxtyping but I have not upgraded that library yet. See here for a discussion (https://github.com/google/jaxtyping/issues/51#issuecomment-1356790303).  Anyway, in the meantime, I just had to punt on checking the output dimension so I use ... for the output dimension.

setup.py
corwinjoy: I think we can do this. It's a bit tricky because the signatures all use Float from jaxtyping but I believe this is all optional stuff. I have moved this to extras_require in setup.py.

test/operators/test_added_diag_linear_operator.py
corwinjoy: The basic idea is that we apply run time tests to libraries imported inside the with block. 
(See https://github.com/google/jaxtyping/blob/main/API.md#option-2-jaxtypinginstall_import_hook).
So, we don't try to test everything all at once which is overwhelming. Instead, this aims to test just the signatures of the code associated with the unit test.  In this case, the with hook will install verification hooks into all the methods in `linear_operator.operators` and then run the tests.

typing_examples.txt
corwinjoy: No. I will kill it. I was more exploring jaxtyping here.

typing_examples_jax.txt
corwinjoy: Same.

linear_operator/test/linear_operator_test_case.py
corwinjoy: I was having trouble pinning down a function signature error and temporarily added exception details. Meant to remove this. Reverted.

linear_operator/test/type_checking_test_case.py
corwinjoy: Thanks!

propagate_type_hints.py
corwinjoy: Agree. This is kind of an ugly location for that file. I wanted to make this obvious but maybe another location would be better.

corwinjoy:
@gpleiss Thanks for taking the time to look at this! I hope that some of these ideas can help to make the library better and make it easier to spot shape inconsistencies. Would still love to add logic to test for unexpected matrix instantiations but I wanted to get the basics first!

corwinjoy:
Note. Also updated code to merge in latest updates to main branch.

setup.py
gpleiss: Never mind... now that I've played around with it, it seems like it would be a pain to make this an optional dependency!

gpleiss:
I played around with this PR a bit today, and I'm super excited about it! A few things I added:

1. I made sure that we are Python 3.8 compatible (a few small sacrifices)
2. I updated our auto-annotate script for the docs to convert the jaxtypes into nice docstring annotations.

Some issues that have come up:

1. Rather than using "..." when we have brodcastable functions, I think we should use `*#batch` (https://github.com/google/jaxtyping/blob/main/API.md#shape). For example:

```python
    def matmul(
        self: Float[LinearOperator, "*#batch M N"],
        other: Union[Float[Tensor, "*#batch N P"], Float[Tensor, " N"], Float[LinearOperator, "*#batch N P"]],
    ) -> Union[Float[Tensor, "*#batch M P"], Float[Tensor, "*#batch M"], Float[LinearOperator, "*#batch M P"]]:
```

This code snippet passes the jaxtype dynamic tests.

2. *Major issue*. The jaxtyping tests only seem to work when I run an individual test file. Currently, the test suite passes if I run `python -m unittest`. However, if I run `python -m unittest test.operators.test_dense_linear_operator` I get several errors... @corwinjoy would you be able to look into this?

gpleiss:
(I just pushed the changes that I played around with today: Python 3.8 compatibility + docstring formatting)

linear_operator/operators/_linear_operator.py
```diff
@@ -6,11 +6,18 @@
...
 from copy import deepcopy
-from typing import Any, Callable, Dict, Optional, Tuple, Union
+from typing import Any, Callable, Dict, Iterable, Optional, Tuple, Union
 
+import numpy as np
 import torch
+
+try:
+    # optional library for advanced type signatures
+    from jaxtyping import Float
```
gpleiss: Since jaxtyping is going to be a non-optional dependency, we shouldn't need this try/except block.

Balandat:
Do we have a sense of the runtime impact of this? I assume it's pretty small and very much negligible for larger matrix sizes, but since this is runtime type checking it would be good to see some benchmarks of how this affects runtime on smaller data sizes.

corwinjoy:
@Balandat There is no runtime impact for the type checking under normal usage. The runtime type checking is an entirely optional feature that can be attached on demand. In the tests we attach type checking via e.g.:
```
from jaxtyping import install_import_hook
with install_import_hook("linear_operator", ("typeguard", "typechecked")):
```

The import_hook adds decorators to the imported functions to perform type checking. The checks should be size-independent since they are looking at the dimension attributes.

linear_operator/operators/_linear_operator.py
corwinjoy: Agreed.

corwinjoy:
@gpleiss Regarding batch shape. I am open to the idea of using `*#batch` but from what I have seen of the linear operator this does not do what we want. As I understand it `*#batch` represents 3 possible cases
1. The dimension is entirely missing. `*` represents 0 or more dimensions.
2. `batch` is present as a dimension and has a value of N.
3. `batch` is present as a dimension and has a value of 1. (This comes from the `#` operator).

Now, I may be wrong, but from what I have seen in the code either the `batch` dimension is present or it is entirely absent. I haven't really seen cases where we leave around broadcastable placeholders with a size of 1, instead, it seems we collapse them. So, I don't think we actually want `*#batch`. Instead, I do believe we want `*batch` where the batch dimension is either present or not.

Balandat:
> There is no runtime impact for the type checking under normal usage. 

Most excellent.

corwinjoy:
@gpleiss Regarding individual unit tests failing.  It looks like I have some more work to do here. It seems that when running the overall unit tests some files may get imported without type checking attached and then signature checking is not enforced. So, instead of adding import_hook into the code, it seems that the correct way is to simply specify runtime type testing as an option when testing. (See https://github.com/google/jaxtyping/blob/main/API.md#pytest-hook).  So, I have added 
```
--jaxtyping-packages=linear_operator,typeguard.typechecked
```
to the pytest parameters via pytest.ini and am going through additional signature issues that have been flagged. (pytest can manage unittest entries, I don't know of a way to pass this argument to unittest directly).
Anyway, this will have the benefit of cleaning up the test code since I can pull out the hooks I added and just have this as a test configuration which will be nice.

CONTRIBUTING.md
corwinjoy: Yes. The type hints are double checked at runtime by using jaxtyping and typeguard.  The context manager installs a hook to add these checks to the routines called by the test suite. It can then verify that the signatures are working as promised. That is, do the type hints match what *actually* happens when the function is called. You can also do static typing with mypy, flake8 etc since jaxtyping is compatible with that. But, these static checks are less accurate.  So, you can do all the things you did before, this just adds optional extra checks.

corwinjoy:
OK. Progress. I changed the code so that when the tests are run the hooks are loaded by pytest itself. This makes sure all functions are checked and the code now looks a lot cleaner. To run tests with type-checking at the top level you can say:
`pytest -c test/typecheck-pytest.ini`
This will run the unit tests with additional run-time type checking.
There is a limitation, for now the typecheck-pytest.ini configuration does not run the kronecker tests since these take too long and overload the stack. I think that the newest version of typeguard will fix this since it goes a lot faster, but they currently have some problems with property getters and setters so that may have to wait a few weeks.
Anyway, all the signatures and tests come back clean so I think this looks pretty good at this point.

I would say this is ready to go, we would need to discuss how we want the automated tests to run.

test/operators/test_added_diag_linear_operator.py
corwinjoy: Update, now we do test everything with the pytest flag. The pro is that this cleans up the tests. The con is that the tests are now a bit slower and the kronecker tests have trouble. But, this does do a more thorough check of all the signature variations. See below for details.

Balandat:
When running this inside a larger code base this throws a bunch of flake8 F722 syntax errors:

    Warning  (FLAKE8) F722
    syntax error in forward annotation '*batch N'
    See https://www.flake8rules.com/rules/F722.html

              61 @typechecker
              62 def linop_matmul_fn_bad_retn(
              63     lo: Float[LinearOperator, "*batch M N"],
    >>>       64     vec: Union[Float[torch.Tensor, "*batch N C"], Float[torch.Tensor, "*batch N"]],
              65 ) -> Float[torch.Tensor, "*batch M C"]:
              66     r"""
              67     As above, but with bad return size

Is this meant to be compatible with flake8? Do I manually need to suppress this error (I don't think we do that in this repo - https://github.com/cornellius-gp/linear_operator/blob/main/setup.cfg#L6).

corwinjoy:
This error is spurious and expected so it can safely be disabled in the
flake8 configuration file. See the following faq from jaxtyping:
https://github.com/google/jaxtyping/blob/main/FAQ.md#flake8-or-ruff-are-throwing-an-error

gpleiss:
@corwinjoy sorry for the slow followup.

I think I have a different understanding of `*#batch`, and also the behavior that we want to describe. Consider `_matmul`, which os currently written as:

```python
    def _matmul(
         self: Float[LinearOperator, "*#batch M N"],
         rhs: Float[torch.Tensor, "*#batch2 N C"],
     ) -> Float[torch.Tensor, "... M C"]:
```

but I think should be written as

```python
    def _matmul(
         self: Float[LinearOperator, "*#batch M N"],
         rhs: Float[torch.Tensor, "*#batch N C"],
     ) -> Float[torch.Tensor, "*#batch M C"]:
```

(Note that my version does seem to pass jaxtyping.) Ultimately, we don't care about the batch shape for `self`, `rhs`, or the return value; as long as they broadcast together.

> Now, I may be wrong, but from what I have seen in the code either the batch dimension is present or it is entirely absent. I haven't really seen cases where we leave around broadcastable placeholders with a size of 1, instead, it seems we collapse them.

There are some cases in GPyTorch where we do rely upon broadcasting operations - e.g. where we wish to multiply a `[B1 x 1 x N x C]` and a `[B2 X M x C]` matrix together. This scenario is even something we test against in our test suite: https://github.com/cornellius-gp/linear_operator/blob/main/linear_operator/test/linear_operator_test_case.py#L368

setup.py
```diff
@@ -37,7 +37,7 @@
     pass
 
 # Other requirements
-install_requires += ["scipy"]
+install_requires += ["scipy", "jaxtyping>=0.2.9", "typeguard~=2.13.3"]
```
gpleiss: Typeguard only needs to be required for tests, right?

Also, could you:
1. Modify the CI files (in the `.github` folder) to run `pytest -c test/typecheck-pytest.ini`
2. Modify the contribution guidelines to mention that `pytest -c test/typecheck-pytest.ini` is the way to run the test suite (rather than `python -m unittest`)

gpleiss: On second thought, maybe it makes sense to get rid of `test/typecheck-pytest.ini`, and we can simply document that you should run purest with the `--jaxtyping-packages=linear_operator,typeguard.typechecked` flags if you want to do type checking.

gpleiss:
After running `pytest test --jaxtyping-packages=linear_operator,typeguard.typechecked`, I'm worries about runtime.

- Running `pytest test` takes ~20s on my machine.
- Running `pytest test --jaxtyping-packages=linear_operator,typeguard.typechecked` takes ~6 minutes!

Not sure what the best solution here is... perhaps we could run type checking on a single LinearOperator (e.g. DenseLinearOperator) since the API is essentially the same for all linear operators and we are propagating the types to all of the subclasses anyways.

@Balandat any thoughts?

corwinjoy:
@gpleiss No problem and thanks for getting back to me.  Let me see if I can explain, and apologies for any confusion. Honestly, I didn't completely understand how Jaxtyping does things until I had traced a few signature verifications. Suppose that for_matmul I have

```python
def _matmul(
        self: Float[LinearOperator, "*batch1 M N"],
        rhs: Float[torch.Tensor, "*batch2 N C"],
    ) -> Float[torch.Tensor, "... M C"]:
```

When passing in matricies the cases I have seen include:
self = `[M x N]`, rhs =  `[N x C]` , return = `[M x C]`
self = `[B1 x M x N]`, rhs =  `[N x C]` , return = `[B1 x M x C]`
self = `[M x N]`, rhs =  `[B1 x N x C]` , return = `[B1 x M x C]`

So, what I *want* to describe with a signature is that the batch dimensions may occur either in the self argument
or the rhs argument and then this batch argument will show up in the output.  
Unfortunately, I wasn't able to really express this in Jaxtyping. Instead, what the above signature says is:
"self *may* have a batch argument (or empty) - assign this to the variable batch1"
"rhs *may* have a batch argument (or empty) - assign this to the variable batch2"
"return *may* include a batch dimension of some kind, do not check it for now"

I had to use two variables here, batch1 and batch2. The reason is, that once Jaxtyping assigns a value to a variable in an argument parameter, for subsequent arguments it just checks that this value matches.  So, I need a batch1 and batch2 since these dimensions are optional / may be different.  I don't use the `#` operator here because this just says "values are the same or must be 1". So

> ```python
>     def _matmul(
>          self: Float[LinearOperator, "*#batch M N"],
>          rhs: Float[torch.Tensor, "*#batch N C"],
>      ) -> Float[torch.Tensor, "*#batch M C"]:
> ```

would fail on 
self = `[B1 x M x N]`, rhs =  `[N x C]` , return = `[B1 x M x C]`

because the variable `batch` would get assigned the value `B1` and then rhs must either have the same batch dimension or 1. i.e. it would require rhs = `[B1 x N x C]` or rhs = `[1 x N x C]`.  And, this is not how rhs is expressed in practice.

At any rate, I agree that the current specification is more lenient than we probably want, but I was unable to figure out how to be stricter since often dimensions are dropped.

Corwin
> 
> I think I have a different understanding of `*#batch`, and also the behavior that we want to describe. Consider `_matmul`, which os currently written as:
> 
> ```python
>     def _matmul(
>          self: Float[LinearOperator, "*#batch M N"],
>          rhs: Float[torch.Tensor, "*#batch2 N C"],
>      ) -> Float[torch.Tensor, "... M C"]:
> ```
> 
> but I think should be written as
> 
> ```python
>     def _matmul(
>          self: Float[LinearOperator, "*#batch M N"],
>          rhs: Float[torch.Tensor, "*#batch N C"],
>      ) -> Float[torch.Tensor, "*#batch M C"]:
> ```
> 
> (Note that my version does seem to pass jaxtyping.) Ultimately, we don't care about the batch shape for `self`, `rhs`, or the return value; as long as they broadcast together.
> 
> > Now, I may be wrong, but from what I have seen in the code either the batch dimension is present or it is entirely absent. I haven't really seen cases where we leave around broadcastable placeholders with a size of 1, instead, it seems we collapse them.
> 
> There are some cases in GPyTorch where we do rely upon broadcasting operations - e.g. where we wish to multiply a `[B1 x 1 x N x C]` and a `[B2 X M x C]` matrix together. This scenario is even something we test against in our test suite: https://github.com/cornellius-gp/linear_operator/blob/main/linear_operator/test/linear_operator_test_case.py#L368

gpleiss:
@corwinjoy thanks for the detailed explanation! What you currently have works for me then :)

setup.py
corwinjoy: > @corwinjoy thanks for the detailed explanation! What you currently have works for me then :)

Cool! I think the short answer was that essentially Jaxtyping has a slightly different notion of broadcasting than LinearOperator. Essentially, Jaxtyping still wants the dimension to be there for broadcasting, but it may be 1. In contrast, LinearOperator will often squeeze / collapse dimensions.
Eventually, I think it would help to be more strict about how these routines are called since then we could have tighter contracts around the signatures. For now, I just tried to reflect the call variation that is present in the tests. For a similar perspective see the following paper which tries to automate this analysis:
https://arxiv.org/pdf/2112.09037.pdf

corwinjoy:
I like the idea of just running typeguard runtime checks on a single LinearOperator for speed with the CI checks.
The newer versions of typeguard also seem to run a lot faster - this runtime may come down to ~1 minute with them.
But, we can't use them yet (as of released version last week) since they still have some bugs to shake out.

> After running `pytest test --jaxtyping-packages=linear_operator,typeguard.typechecked`, I'm worries about runtime.
> 
> * Running `pytest test` takes ~20s on my machine.
> * Running `pytest test --jaxtyping-packages=linear_operator,typeguard.typechecked` takes ~6 minutes!
> 
> Not sure what the best solution here is... perhaps we could run type checking on a single LinearOperator (e.g. DenseLinearOperator) since the API is essentially the same for all linear operators and we are propagating the types to all of the subclasses anyways.
> 
> @Balandat any thoughts?

Balandat:
> But, we can't use them yet (as of released version last week) since they still have some bugs to shake out.

I think depending on the very latest typeguard may also be problematic, ideally we don't need to live on the bleeding edge (I'm running linear_operator in a massive monorepo, and tasks like upgrading a third party lib dep taht seem simple can turn into massive engineering projects). 

Though if typeguard is only required for running the typing test, I figure I can just not run those and rely on the project CI.

I think running the type checks routinely on a single linear_operator for every commit makes sense. Though we probably want a nightly job that runs the whole codebase as well.

gpleiss (approved):

gpleiss:
Finally got this in! Thanks @corwinjoy !

corwinjoy:
Awesome! Thanks so much for reviewing, polishing and merging this @gpleiss. I know it was a lot to review!

Balandat:
Hmm @gpleiss I might have missed this, but it looks like `jaxtyping` is now a required dep even for installation, rather than just testing? In that case we need to add it to `install_requires`.  Will put up a PR for this

linear_operator/operators/diag_linear_operator.py
```diff
@@ -44,92 +48,115 @@ def _bilinear_derivative(self, left_vecs: Tensor, right_vecs: Tensor) -> Tuple[T
...
+    def _root_inv_decomposition(
+        self: Float[LinearOperator, "*batch N N"],
+        initial_vectors: Optional[torch.Tensor] = None,
+        test_vectors: Optional[torch.Tensor] = None,
+    ) -> Union[Float[LinearOperator, "... N N"], Float[Tensor, "... N N"]]:
         return self.inverse().sqrt()
 
     def _size(self) -> torch.Size:
-        return self._diag.shape + self._diag.shape[-1:]
+        return torch.Size([*self._diag.shape, self._diag.shape[-1]])
```
Balandat, line 105: This change broke the case where `self._diag` is a scalar tensor. Will put up a fix

PR #43, opened 2022-12-16
Cleaned up configuration for isort and black

JonathanWenger:
# In a Nutshell
Configuration between linters `isort` and `black` was only handled in pre-commit hooks and could lead to conflicts.

# In Detail
When running pre-commit hooks and also locally installing black via the instructions in `CONTRIBUTING.md` one could end up with different versions / settings leading to reformatting from the pre-commit hooks every commit. I also fixed the exclusions in the pre-commit configuration along the way, since it was throwing warnings:

```
 ▶ git commit -m "foo"
[WARNING] The 'exclude' field in hook 'flake8' is a regex, not a glob -- matching '/*' probably isn't what you want here
[WARNING] The 'exclude' field in hook 'black' is a regex, not a glob -- matching '/*' probably isn't what you want here
[WARNING] The 'exclude' field in hook 'isort' is a regex, not a glob -- matching '/*' probably isn't what you want here
[WARNING] The 'exclude' field in hook 'require-ascii' is a regex, not a glob -- matching '/*' probably isn't what you want here
[WARNING] The 'exclude' field in hook 'forbid-binary' is a regex, not a glob -- matching '/*' probably isn't what you want here
```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -16,26 +16,38 @@ repos:
     hooks:
     -   id: flake8
         args: [--config=setup.cfg]
-        exclude: ^(examples/*)|(docs/*)
+        exclude: |
+            ^examples/
+            ^docs/
 -   repo: https://github.com/ambv/black
     rev: 22.3.0
     hooks:
     -   id: black
-        exclude: ^(build/*)|(docs/*)|(examples/*)
+        exclude: |
+            ^build/
+            ^docs/
+            ^examples/
 -   repo: https://github.com/pre-commit/mirrors-isort
     rev: v5.10.1
     hooks:
     -   id: isort
         language_version: python3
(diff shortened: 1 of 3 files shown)


JonathanWenger:
It seems like there are some inconsistencies in the import formatting in the current library. Let me know if you're fine with me just reformatting the imports in this PR.

Balandat:
How do you (and @gpleiss, @jacobrgardner) feel about using µfmt (https://ufmt.omnilib.dev/en/stable/)? This takes care of the conflicts between black and isort, we've been using it without issues in BoTorch for a long time now.

gpleiss:
I am 100% in favor of any dev practices you all use for BoTorch :)

Balandat:
Done in #56.

Issue #44, opened 2023-01-02
[Bug] Passing `other: Tensor` to `LinearOperator.to`.


j-wilson:
# 🐛 Bug

`torch.Tensor.to` allows the user to provide a reference tensor in lieu of explicitly passing `device` and `dtype`; however, this logic seems to be missing from `LinearOperator.to`.

## To reproduce
```
A = DiagLinearOperator(torch.rand(3, dtype=torch.float64))
B = A.to(torch.rand(3, dtype=torch.float32))
B.dtype
> torch.float64
```

## Expected Behavior

The code should either follow the same convention as `torch.Tensor.to` or raise an exception.

PR #45, opened 2023-01-02
Allow `.to()` to operate on tensors

Balandat:
Addresses #44.

### DIFF
--- a/linear_operator/utils/generic.py
+++ b/linear_operator/utils/generic.py
@@ -44,7 +44,7 @@ def _to_helper(*args, **kwargs) -> Tuple[Optional[torch.device], Optional[torch.
     if len(device_args) > 1:
         raise RuntimeError(f"{base_error_message} devices ({device_args}.")
 
-    dtype = next(iter(dtype_args)) if dtype_args else None
-    device = next(iter(device_args)) if device_args else None
+    dtype = dtype_args.pop() if dtype_args else None
+    device = device_args.pop() if device_args else None
 
     return device, dtype
--- a/linear_operator/utils/generic.py
+++ b/linear_operator/utils/generic.py
@@ -1,0 +1,50 @@
+#!/usr/bin/env python3
+
+from __future__ import annotations
+
+from typing import Optional, Set, Tuple
+
+import torch
+
+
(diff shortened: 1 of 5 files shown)


linear_operator/utils/generic.py
```diff
@@ -0,0 +1,50 @@
...
+        device_args.add(kwargs["device"])
+
+    # Handle the case when casting is ambiguous
+    base_error_message = "Attempted to cast LinearOperator object to multiple"
+    if len(dtype_args) > 1:
+        raise RuntimeError(f"{base_error_message} dtypes ({dtype_args}.")
+    if len(device_args) > 1:
+        raise RuntimeError(f"{base_error_message} devices ({device_args}.")
+
+    dtype = next(iter(dtype_args)) if dtype_args else None
```
SebastianAment:
```suggestion
    dtype = dtype_args.pop() if dtype_args else None
```

linear_operator/utils/generic.py
```diff
@@ -0,0 +1,50 @@
...
+    device = next(iter(device_args)) if device_args else None
```
SebastianAment:
```suggestion
    device = device_args.pop() if device_args else None
```

test/utils/test_generic.py
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env python3
+
+import unittest
+
+import torch
+
+from linear_operator.test.base_test_case import BaseTestCase
+from linear_operator.utils.generic import _to_helper
+
+
```
SebastianAment, line 10: In order to test that this does the right thing for non-cpu tensors, could add a mock tensor like this:
```suggestion
class TestTensor(Tensor):
    @property
    def device(self):
        return self._device
```
Then in the test do 
```python
A = TestTensor()
A._device = torch.device("cuda")
```
`A` will then pretend to be on a CUDA device, without us having to have access to one.

SebastianAment (approved):
LGTM! Left comments inline. Could be worth adding the non-cpu tensor test before merging.

test/utils/test_generic.py
Balandat: Yeah makes sense, or I can just use `mock` directly to assign the attribute.

Issue #46, opened 2023-01-06
argument name inconsistency in setting jitter value with gpytorch


LuhuanWu:
- Change the init argument of class `_dtype_value_context`. 

    - Currently the init function for `_dtype_value_context` in [`linear_operator/settings.py`](https://github.com/cornellius-gp/linear_operator/blob/71c199ae87a965e6591c8ea3401583450f9ad0a2/linear_operator/settings.py#L36)  is: 
    ```def __init__(self, float=None, double=None, half=None):```

    - This is not consistent with the init function for  `_dtype_value_context` in [`gpytorch/settings.py`](https://github.com/cornellius-gp/gpytorch/blob/0fc5408e0aa98b7fa6705e4d21c13068b4e5b1b6/gpytorch/settings.py#L59), which is: 
     ```def __init__(self, float_value=None, double_value=None, half_value=None):``` 

    - I would suggest to unify the argument names, for example, make the both init functions as:
    ```def __init__(self, float_value=None, double_value=None, half_value=None):```

Balandat:
Thanks for raising this. I think the args without the `_value` suffix are clear enough and much shorter, so maybe we should make this change on the gyptorch end. I can put up a quick PR. that is backward compatible.

Balandat:
Actually that doesn't make a lot of sense since it will override the python standard `float` type in that function. Makes more sense to make the change you suggested.

Balandat:
That makes sense to me. Will put up a PR

gpleiss:
Should be closed by #47

PR #47, opened 2023-01-11
Change _dtype_value_context init args to `*_value`.

Balandat:
Addresses #46. Also gets rid of an old `DeprecationWarning`


gpleiss (approved):

saitcakmak:
This is missing a `dtype` arg in `self.__class__.value()` calls. Working on a fix.

Surfaced in https://github.com/pytorch/botorch/actions/runs/3990958264/jobs/6845226501

PR #49, opened 2023-01-24
Fix missing dtype arg in _dtype_value_context 

saitcakmak:
This was missing the `dtype` arg in `self.__class__.value(...)` call, leading to failures like https://github.com/pytorch/botorch/actions/runs/3990958264/jobs/6845226501.

Added (limited) units to help catch similar errors in the future.


Balandat (approved):
Thanks - someone must have been a bit trigger happy with the merge of my previous change :)

PR #51, opened 2023-02-22
Remove @cached decorator for TriangularLinearOperator.to_dense

Balandat:
This seems unnecessary and also causes issues with pickling: https://github.com/cornellius-gp/gpytorch/issues/2264#issue-1567345205


esantorella:
My gut instinct is that this is right, since I've seen the same dense tensor getting cached to multiple places, including via this function. If this causes any performance degradation, hopefully the BoTorch tutorials runs and the tracking I'm adding to them would pick that up (https://github.com/pytorch/botorch/pull/1695).

gpleiss (approved):

PR #56, opened 2023-03-30
Format using ufmt, upgrade pre-commit config

Balandat:
This PR introduces the us of ufmt (instead of black + isort) for auto-formatting and import sorting to avoid collisions between the two. It also upgrades some of the versions (flake8 -> 5.0.4, flake8-print -> 5.0.0, pre-commit-hooks -> 4.4.0). As a result, quire a few files were reformatted.

### DIFF
--- a/linear_operator/utils/getitem.py
+++ b/linear_operator/utils/getitem.py
@@ -14,7 +14,8 @@ _noop_index = slice(None, None, None)
 
 
 def _compute_getitem_size(
-    obj: Union[torch.Tensor, "LinearOperator"], indices: Tuple[Union[slice, torch.LongTensor, int], ...]  # noqa F811
+    obj: Union[torch.Tensor, "LinearOperator"],  # noqa F821
+    indices: Tuple[Union[slice, torch.LongTensor, int], ...],  # noqa F811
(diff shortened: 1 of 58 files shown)


linear_operator/utils/getitem.py
```diff
@@ -14,7 +14,8 @@
 
 
 def _compute_getitem_size(
-    obj: Union[torch.Tensor, "LinearOperator"], indices: Tuple[Union[slice, torch.LongTensor, int], ...]  # noqa F811
+    obj: Union[torch.Tensor, "LinearOperator"],  # noqa F821
```
Balandat, line 17: I had to silence what I think is a spurious flake8 error here, this complained about `LinearOperator` not being known / imported ...

gpleiss:
I'm very much in favor of this. I'm going to hopefully get #42 in early next week and then we can merge this.

Balandat:
> I'm very much in favor of this. I'm going to hopefully get https://github.com/cornellius-gp/linear_operator/pull/42 in early next week and then we can merge this.

The reason for this is avoiding a not-so-nice rebase I take it?

gpleiss (approved):

PR #60, opened 2023-05-03
Make jaxtyping dep install requirement

Balandat:
Make jaxtyping and typeguard required dependencies

See https://github.com/cornellius-gp/linear_operator/pull/42#issuecomment-1533225278

Also removes the `packaging` dep, which is transitive through `pytest`: https://github.com/pytest-dev/pytest/blob/main/setup.cfg#L48

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -77,7 +77,7 @@ setup(
     install_requires=install_requires,
     extras_require={
         "dev": ["ufmt", "twine", "pre-commit"],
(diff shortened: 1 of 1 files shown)


Balandat:
Hmm @gpleiss looks like `typeguard` is a required dep of `jaxtyping`. We've had issues with running `typeguard` on internal infra, so this isn't great. Is there a way to make `jaxtyping` truly a test-time requirement only as we initially intended with #42?

Balandat:
cc @mpolson64

Balandat:
Merging this to resolve breakage of dev version install.

PR #62, opened 2023-05-05
Add KernelLinearOperator, deprecate KeOpsLinearOperator

gpleiss:
KeOpsLinearOperator does not correctly backpropagate gradients if the covar_func closes over parameters.

KernelLinearOperator corrects for this, and is set up to replace LazyEvaluatedKernelTensor in GPyTorch down the line.

[Addresses issues in [#2296](https://github.com/cornellius-gp/gpytorch/pull/2296#issuecomment-1487613562)]

### DIFF
--- a/linear_operator/operators/_linear_operator.py
+++ b/linear_operator/operators/_linear_operator.py
@@ -375,7 +375,8 @@ class LinearOperator(object):
         with torch.autograd.enable_grad():
             lin_op = self.representation_tree()(*args)
             loss = (left_vecs * lin_op._matmul(right_vecs)).sum()
-            actual_grads = deque(torch.autograd.grad(loss, args, allow_unused=True))
+            args_with_grads = [arg for arg in args if arg.requires_grad]
+            actual_grads = deque(torch.autograd.grad(loss, args_with_grads, allow_unused=True))
 
         # Now make sure that the object we return has one entry for every item in args
         grads = []
--- a/linear_operator/operators/_linear_operator.py
+++ b/linear_operator/operators/_linear_operator.py
@@ -2042,7 +2042,7 @@ class LinearOperator(object):
         Returns the Tensors that are used to define the LinearOperator
         """
         representation = []
-        for arg in list(self._args) + list(self._differentiable_kwarg_vals):
+        for arg in itertools.chain(self._args, self._differentiable_kwarg_vals):
             if torch.is_tensor(arg):
                 representation.append(arg)
             elif hasattr(arg, "representation") and callable(arg.representation):  # Is it a LinearOperator?
--- a/linear_operator/operators/_linear_operator.py
+++ b/linear_operator/operators/_linear_operator.py
@@ -367,6 +367,10 @@ class LinearOperator(object):
             else:
                 args.append(arg.detach())
 
+        # If no arguments require gradients, then we're done!
+        if not any(arg.requires_grad for arg in args):
+            return (None,) * len(args)
+
         # We'll use the autograd to get us a derivative
         with torch.autograd.enable_grad():
             lin_op = self.representation_tree()(*args)
--- a/linear_operator/operators/_linear_operator.py
+++ b/linear_operator/operators/_linear_operator.py
@@ -350,18 +350,19 @@ class LinearOperator(object):
         """
         from collections import deque
 
-        args = tuple(self.representation())
-        args_with_grads = tuple(arg for arg in args if arg.requires_grad)
+        # Construct a detached version of each argument in the linear operator
+        args = []
+        for arg in self.representation():
+            if torch.is_tensor(arg) and arg.dtype.is_floating_point:
+                args.append(arg.detach().requires_grad_(True))
+            else:
+                args.append(arg.detach())
 
-        # Easy case: if we don't require any gradients, then just return!
-        if not len(args_with_grads):
-            return tuple(None for _ in args)
-
-        # Normal case: we'll use the autograd to get us a derivative
+        # We'll use the autograd to get us a derivative
         with torch.autograd.enable_grad():
-            loss = (left_vecs * self._matmul(right_vecs)).sum()
-            loss.requires_grad_(True)
-            actual_grads = deque(torch.autograd.grad(loss, args_with_grads, allow_unused=True))
+            lin_op = self.representation_tree()(*args)
+            loss = (left_vecs * lin_op._matmul(right_vecs)).sum()
+            actual_grads = deque(torch.autograd.grad(loss, args, allow_unused=True))
 
         # Now make sure that the object we return has one entry for every item in args
         grads = []
@@ -1344,7 +1345,11 @@ class LinearOperator(object):
         (In practice, this function removes all Tensors that make up the
         :obj:`~linear_operator.opeators.LinearOperator` from the computation graph.)
         """
-        return self.clone().detach_()
+        detached_args = [arg.detach() if hasattr(arg, "detach") else arg for arg in self._args]
+        detached_kwargs = dict(
+            (key, val.detach() if hasattr(val, "detach") else val) for key, val in self._kwargs.items()
+        )
+        return self.__class__(*detached_args, **detached_kwargs)
 
     def detach_(self: Float[LinearOperator, "*batch M N"]) -> Float[LinearOperator, "*batch M N"]:
         """
--- a/linear_operator/operators/_linear_operator.py
+++ b/linear_operator/operators/_linear_operator.py
@@ -3,10 +3,12 @@
 from __future__ import annotations
 
 import functools
+import itertools
 import math
 import numbers
 import warnings
 from abc import abstractmethod
+from collections import OrderedDict
 from copy import deepcopy
 from typing import Any, Callable, Dict, List, Optional, Tuple, Union
 
@@ -150,14 +152,12 @@ class LinearOperator(object):
                 raise ValueError(err)
 
         self._args = args
-        self._differentiable_kwarg_names = []
-        self._differentiable_kwarg_vals = []
+        self._differentiable_kwargs = OrderedDict()
         self._nondifferentiable_kwargs = dict()
         for name, val in sorted(kwargs.items()):
             # Sorting is necessary so that the flattening in the representation tree is deterministic
             if torch.is_tensor(val) or isinstance(val, LinearOperator):
-                self._differentiable_kwarg_names.append(name)
-                self._differentiable_kwarg_vals.append(val)
+                self._differentiable_kwargs[name] = val
             else:
                 self._nondifferentiable_kwargs[name] = val
 
@@ -362,7 +362,8 @@ class LinearOperator(object):
         # Construct a detached version of each argument in the linear operator
         args = []
         for arg in self.representation():
-            if torch.is_tensor(arg) and arg.dtype.is_floating_point and arg.requires_grad:
+            # All arguments here are guaranteed to be tensors
+            if arg.dtype.is_floating_point and arg.requires_grad:
                 args.append(arg.detach().requires_grad_(True))
             else:
                 args.append(arg.detach())
@@ -473,14 +474,8 @@ class LinearOperator(object):
         self._args_memo = args
 
     @property
-    def _differentiable_kwargs(self) -> Dict[str, Union[Tensor, "LinearOperator"]]:
-        return dict(zip(self._differentiable_kwarg_names, self._differentiable_kwarg_vals))
-
-    @property
     def _kwargs(self) -> Dict[str, Any]:
-        kwargs = self._differentiable_kwargs
-        kwargs.update(self._nondifferentiable_kwargs)
-        return kwargs
+        return {**self._differentiable_kwargs, **self._nondifferentiable_kwargs}
 
     def _approx_diagonal(self: Float[LinearOperator, "*batch N N"]) -> Float[torch.Tensor, "*batch N"]:
         """
@@ -2042,7 +2037,7 @@ class LinearOperator(object):
         Returns the Tensors that are used to define the LinearOperator
         """
         representation = []
-        for arg in itertools.chain(self._args, self._differentiable_kwarg_vals):
+        for arg in itertools.chain(self._args, self._differentiable_kwargs.values()):
             if torch.is_tensor(arg):
                 representation.append(arg)
             elif hasattr(arg, "representation") and callable(arg.representation):  # Is it a LinearOperator?
--- a/linear_operator/operators/_linear_operator.py
+++ b/linear_operator/operators/_linear_operator.py
@@ -362,7 +362,7 @@ class LinearOperator(object):
         # Construct a detached version of each argument in the linear operator
         args = []
         for arg in self.representation():
-            if torch.is_tensor(arg) and arg.dtype.is_floating_point:
+            if torch.is_tensor(arg) and arg.dtype.is_floating_point and arg.requires_grad:
                 args.append(arg.detach().requires_grad_(True))
             else:
                 args.append(arg.detach())
@@ -468,10 +468,13 @@ class LinearOperator(object):
         self._args_memo = args
 
     @property
+    def _differentiable_kwargs(self) -> Dict[str, Union[Tensor, "LinearOperator"]]:
+        return dict(zip(self._differentiable_kwarg_names, self._differentiable_kwarg_vals))
+
+    @property
     def _kwargs(self) -> Dict[str, Any]:
-        kwargs = dict(
-            zip(self._differentiable_kwarg_names, self._differentiable_kwarg_vals), **self._nondifferentiable_kwargs
-        )
+        kwargs = self._differentiable_kwargs
+        kwargs.update(self._nondifferentiable_kwargs)
         return kwargs
 
     def _approx_diagonal(self: Float[LinearOperator, "*batch N N"]) -> Float[torch.Tensor, "*batch N"]:
--- a/linear_operator/operators/_linear_operator.py
+++ b/linear_operator/operators/_linear_operator.py
@@ -150,7 +150,16 @@ class LinearOperator(object):
                 raise ValueError(err)
 
         self._args = args
-        self._kwargs = kwargs
+        self._differentiable_kwarg_names = []
+        self._differentiable_kwarg_vals = []
+        self._nondifferentiable_kwargs = dict()
+        for name, val in sorted(kwargs.items()):
+            # Sorting is necessary so that the flattening in the representation tree is deterministic
+            if torch.is_tensor(val) or isinstance(val, LinearOperator):
+                self._differentiable_kwarg_names.append(name)
+                self._differentiable_kwarg_vals.append(val)
+            else:
+                self._nondifferentiable_kwargs[name] = val
 
     ####
     # The following methods need to be defined by the LinearOperator
@@ -457,6 +466,13 @@ class LinearOperator(object):
     @_args.setter
     def _args(self, args: Tuple[Union[torch.Tensor, "LinearOperator", int], ...]) -> None:
         self._args_memo = args
+
+    @property
+    def _kwargs(self) -> Dict[str, Any]:
+        kwargs = dict(
+            zip(self._differentiable_kwarg_names, self._differentiable_kwarg_vals), **self._nondifferentiable_kwargs
+        )
+        return kwargs
 
     def _approx_diagonal(self: Float[LinearOperator, "*batch N N"]) -> Float[torch.Tensor, "*batch N"]:
         """
@@ -2018,7 +2034,7 @@ class LinearOperator(object):
         Returns the Tensors that are used to define the LinearOperator
         """
         representation = []
-        for arg in self._args:
+        for arg in list(self._args) + list(self._differentiable_kwarg_vals):
             if torch.is_tensor(arg):
                 representation.append(arg)
             elif hasattr(arg, "representation") and callable(arg.representation):  # Is it a LinearOperator?
--- a/linear_operator/operators/kernel_linear_operator.py
+++ b/linear_operator/operators/kernel_linear_operator.py
@@ -6,6 +6,7 @@ import torch
 from jaxtyping import Float
 from torch import Tensor
 
+from ..utils.broadcasting import _pad_with_singletons
 from ..utils.getitem import _noop_index, IndexType
 from ..utils.memoize import cached
 from ._linear_operator import LinearOperator, to_dense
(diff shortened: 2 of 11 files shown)


gpleiss:
@jacobrgardner @Balandat can one of you take a look at this?

m-julian:
This implementation works perfectly when using one KeOps kernel, but there are a few problems when combining kernels. This is not specifically a problem with `KernelLinearOperator`, but the way the things are implemented in `gpytorch`/`linear_operator` currently. Hopefully some of these comments are useful for structuring the code.

I tried to multiply two keops kernels (which make a `ProductKernel`). These are the problems I found:
1. `ProductKernel` calls `to_dense` if x1 and x2 are not identical [here](https://github.com/cornellius-gp/gpytorch/blob/c6a4f46838c26e33795cfc30b6e1fd31d72e26e4/gpytorch/kernels/kernel.py#L653-L658) and returns a `DenseLinearOperator` since `MulLinearOperator` is only used for square symmetric matrices currently. Since KeOps can be used for large matrices where x1 is not equal to x2, calling `to_dense` could give memory errors. Instead, this should give back some object that represents the multiplication of the two symbolic matrices without actually computing them.
2. If x1 is equal to x2, then you end up with a `MulLinearOpeartor` which calls `root_decomposition` on both the `KerneLinearOperator`s. Then there are two issues:
2.1. You might run out of memory if the matrix is very large when doing the root decomposition (because of KeOps).
2.2. If the matrix fits into memory, it will be approximated as Lanczos is the default algorithm. This is fixed by doing `gpytorch.settings.fast_computations(covar_root_decomposition=False)`, but it took me a while to figure out why I was getting different results when using KeOps kernels. If using a very large KeOps symbolic matrix, then I don't think root decomposition should be called, but instead the symbolic matrix should be used until some reduction operation is performed. (So should  `MulLinearOperator` be used with `KernelLinearOperator` instances?)

I was thinking could `KernelLinearOperator` subclass from `DenseLinearOperator` since the `KernelLinearOperator` class is used to represent a full covariance matrix (just computed on the fly when a reduction operation is applied)? That way all the checks for `isinstance(self, DenseLinearOperator)` (for example [this one](https://github.com/cornellius-gp/linear_operator/blob/e69133d2aefd9a5276147bb95588ed6c56a637db/linear_operator/operators/_linear_operator.py#L584-L589)) also work for `KernelLinearOperator`. Then this goes around the problem with `MulLinearOperator` but might cause other problems, so not sure if it is reasonable.

linear_operator/operators/_linear_operator.py
```diff
@@ -350,18 +350,19 @@ def _bilinear_derivative(self, left_vecs: Tensor, right_vecs: Tensor) -> Tuple[O
...
-        args_with_grads = tuple(arg for arg in args if arg.requires_grad)
-
-        # Easy case: if we don't require any gradients, then just return!
-        if not len(args_with_grads):
-            return tuple(None for _ in args)
+        # Construct a detached version of each argument in the linear operator
+        args = []
+        for arg in self.representation():
+            if torch.is_tensor(arg) and arg.dtype.is_floating_point:
+                args.append(arg.detach().requires_grad_(True))
```
Balandat, line 367: Should `requires_grad` be set here depending on whether original `arg` has `requires_grad` set (as was the case before)?

linear_operator/operators/kernel_linear_operator.py
```diff
@@ -0,0 +1,217 @@
...
+        lengthscale = torch.randn(2, 1, 1, 6)
+        outputscale = torch.randn(2, 1, 1, 1)
+        kern = KernelLinearOperator(x1, x2, lengthscale, outputscale, covar_func=covar_func)
+
+        # kern is of size 2 x 3 x 5 x 4
+
+    .. warning ::
+
+        `covar_func` should not close over any parameters. Any parameters that are closed over will not have
+        propagated gradients.
```
Balandat, line 114: Will everyone who is working on GPs but may not have any formal CS training know what it means to "close over"? Might be worth spelling that out and given an example.

linear_operator/operators/kernel_linear_operator.py
```diff
@@ -0,0 +1,217 @@
...
+
+    :param x1: The data :math:`\boldsymbol X_1.`
+    :param x2: The data :math:`\boldsymbol X_2.`
+    :param params: Additional hyperparameters (:math:`\boldsymbol \theta`) passed into covar_func.
+    :param covar_func: The covariance function :math:`k_{\boldsymbol \theta}(\cdot, \cdot)`.
+        Its arguments should be `x1`, `x2`, `*params`, `**kwargs`, and it should output the covariance matrix
+        between :math:`\boldsymbol X_1` and :math:`\boldsymbol X_2`.
+    :param kwargs: Any additional (non-hyperparameter) kwargs to pass into `covar_func`.
```
Balandat: It's a bit weird that all hypers are assumed to be positional args here. It seems that this is rather opaque gotcha, and could also result in making the code much less readable.

linear_operator/operators/kernel_linear_operator.py
```diff
@@ -0,0 +1,217 @@
...
+            if isinstance(batch_indices, slice):
+                x2 = self.x2.expand(1, *self.x2.shape[-2:])[(*batch_indices, row_index, dim_index)]
+            elif isinstance(batch_indices, tuple):
+                if any([not isinstance(bi, slice) for bi in batch_indices]):
+                    raise RuntimeError(
+                        "Attempting to tensor index a non-batch matrix's batch dimensions. "
+                        f"Got batch index {batch_indices} but my shape was {self.shape}"
+                    )
+                x2 = self.x2.expand(*([1] * len(batch_indices)), *self.x2.shape[-2:])
+                x2 = x2[(*batch_indices, row_index, dim_index)]
```
Balandat: you'll probably want to throw this into a helper function to deduplicate the code here

test/operators/test_identity_linear_operator.py
```diff
@@ -76,6 +76,10 @@ def create_linear_op(self):
     def evaluate_linear_op(self, linear_op):
         return torch.eye(5)
 
+    def test_bilinear_derivative(self):
+        # Not needed, and it errors for some reason
```
Balandat: Do we know the reason? Might be good to understand

gpleiss:
@m-julian there is unfortunately no way to do a symbolic element-wise multiplication of two kernels. KeOps (and LinearOperator) can keep things symbolic by using matrix multiplication based algorithms (CG/Lanczos) for solves and log determinants. Unfortunately, matrix multiplication does not distribute over element wise product. The current Lanczos solution comes out of the Product Kernel Interpolation paper (it was the best solution we could come up with).

Therefore, I don't know if there's a better way to handle the product of two matrices than what we currently do in code.

> I was thinking could KernelLinearOperator subclass from DenseLinearOperator since the KernelLinearOperator class is used to represent a full covariance matrix (just computed on the fly when a reduction operation is applied)?

This would probably be a can of worms. DenseLinearOperator assumes that the matrix is represented by a tensor. 
And most of our LinearOperator classes represent full covariance matrices that are computed on the fly when reduction operations are called.

linear_operator/operators/kernel_linear_operator.py
gpleiss: @Balandat I can try to allow for kwarg hypers, but this might be a can of worms.

gpleiss: @Balandat okay I changed it so that hypers must be kwargs (cannot be positional args).

This change required modifying some base LinearOperator functionality so that the flattening/unflattening that we do in representation trees (for CG backward passes, etc.) accounts for LinearOperator kwargs as well as args.

linear_operator/operators/_linear_operator.py
gpleiss: Fixed.

linear_operator/operators/kernel_linear_operator.py
gpleiss: Fixed. I referenced the example in the documentation.

test/operators/test_identity_linear_operator.py
gpleiss: I fixed an issue in LinearOperator._bilinear_derivative (it now exits early if no arguments require gradients), so this short-circuiting is no longer necessary.

linear_operator/operators/kernel_linear_operator.py
gpleiss: Done.

Turakar:
Regarding the product discussion: Regarding a general implementation that would be part of GPyTorch, I also do not know of a better approach than what @gpleiss pointed out. However, from a user's perspective, you, @m-julian, could of course write a custom KeOps kernel. Most likely, this would even be faster than two separate kernels, as you only need one trip from global memory to processor registers on the GPU.

linear_operator/operators/_linear_operator.py
```diff
@@ -350,17 +359,23 @@ def _bilinear_derivative(self, left_vecs: Tensor, right_vecs: Tensor) -> Tuple[O
...
 
-        args = tuple(self.representation())
-        args_with_grads = tuple(arg for arg in args if arg.requires_grad)
+        # Construct a detached version of each argument in the linear operator
+        args = []
+        for arg in self.representation():
+            if torch.is_tensor(arg) and arg.dtype.is_floating_point and arg.requires_grad:
+                args.append(arg.detach().requires_grad_(True))
+            else:
+                args.append(arg.detach())
```
Balandat, line 369: may want to comment here that the representation is guaranteed to consist only or tensors

linear_operator/operators/_linear_operator.py
```diff
@@ -150,7 +150,16 @@ def __init__(self, *args, **kwargs):
...
         self._args = args
-        self._kwargs = kwargs
+        self._differentiable_kwarg_names = []
+        self._differentiable_kwarg_vals = []
+        self._nondifferentiable_kwargs = dict()
+        for name, val in sorted(kwargs.items()):
+            # Sorting is necessary so that the flattening in the representation tree is deterministic
+            if torch.is_tensor(val) or isinstance(val, LinearOperator):
+                self._differentiable_kwarg_names.append(name)
+                self._differentiable_kwarg_vals.append(val)
```
Balandat: Why not just use an (ordered) dict for this?

linear_operator/operators/_linear_operator.py
```diff
@@ -457,6 +472,16 @@ def _args(self) -> Tuple[Union[torch.Tensor, "LinearOperator", int], ...]:
     def _args(self, args: Tuple[Union[torch.Tensor, "LinearOperator", int], ...]) -> None:
         self._args_memo = args
 
+    @property
+    def _differentiable_kwargs(self) -> Dict[str, Union[Tensor, "LinearOperator"]]:
```
Balandat: could avoid this by using an ordered dict to store these it seems?

linear_operator/operators/_linear_operator.py
```diff
@@ -457,6 +472,16 @@ def _args(self) -> Tuple[Union[torch.Tensor, "LinearOperator", int], ...]:
...
+        return dict(zip(self._differentiable_kwarg_names, self._differentiable_kwarg_vals))
+
+    @property
+    def _kwargs(self) -> Dict[str, Any]:
+        kwargs = self._differentiable_kwargs
+        kwargs.update(self._nondifferentiable_kwargs)
```
Balandat: hmm this only works b/c you're generating the `_differentiable_kwargs` dynamically as a property. Worth commenting on this here. Alternatively just use a dict to store these, and then do sth like 

    @property
    def _kwargs(self) -> Dict[str, Any]:
        return {**self._differentiable_kwargs, **self._nondifferentiable_kwargs}

here.

linear_operator/operators/_linear_operator.py
```diff
@@ -2013,7 +2042,7 @@ def representation(self) -> Tuple[torch.Tensor, ...]:
         Returns the Tensors that are used to define the LinearOperator
         """
         representation = []
-        for arg in self._args:
+        for arg in list(self._args) + list(self._differentiable_kwarg_vals):
```
Balandat: Can be agnostic to the types here if you just want to iterate
```suggestion
        for arg in itertools.chain(self._args, self._differentiable_kwarg_vals):
```

linear_operator/operators/kernel_linear_operator.py
```diff
@@ -0,0 +1,385 @@
...
+    """
+    try:
+        x = x[(*batch_indices, data_index, _noop_index)]
+    # We're going to handle multi-batch indexing with a try-catch loop
+    # This way - in the default case, we can avoid doing expansions of x1 which can be timely
+    except IndexError:
+        if isinstance(batch_indices, slice):
+            x = x.expand(1, *x.shape[-2:])[(*batch_indices, data_index, _noop_index)]
+        elif isinstance(batch_indices, tuple):
+            if any([not isinstance(bi, slice) for bi in batch_indices]):
```
Balandat:
```suggestion
            if any(not isinstance(bi, slice) for bi in batch_indices):
```

linear_operator/operators/kernel_linear_operator.py
```diff
@@ -0,0 +1,385 @@
...
+    (Sometimes x needs to be expanded before calling x[*batch_indices, data_index, :]; i.e. if
+    the batch_indices broadcast. We try to prevent this expansion if possible.
+    """
+    try:
+        x = x[(*batch_indices, data_index, _noop_index)]
+    # We're going to handle multi-batch indexing with a try-catch loop
+    # This way - in the default case, we can avoid doing expansions of x1 which can be timely
+    except IndexError:
+        if isinstance(batch_indices, slice):
+            x = x.expand(1, *x.shape[-2:])[(*batch_indices, data_index, _noop_index)]
```
Balandat: Readability 
```suggestion
            x = x.expand(1, *x.shape[-2:])
            x = [(*batch_indices, data_index, _noop_index)]
```

linear_operator/operators/kernel_linear_operator.py
```diff
@@ -0,0 +1,385 @@
...
+    :param x1: The data :math:`\boldsymbol X_1.`
+    :param x2: The data :math:`\boldsymbol X_2.`
+    :param covar_func: The covariance function :math:`k_{\boldsymbol \theta}(\cdot, \cdot)`.
+        Its arguments should be `x1`, `x2`, `**params`, and it should output the covariance matrix
+        between :math:`\boldsymbol X_1` and :math:`\boldsymbol X_2`.
+    :param num_outputs_per_input: The number of outputs per data point.
+        This parameter should be 1 for most kernels, but will be >1 for multitask kernels,
+        gradient kernels, and any other kernels that require cross-covariance terms for multiple domains.
+        If a tuple is passed, there will be a different number of outputs per input dimension
+        for the rows/cols of the kernel matrix.
```
Balandat, line 128: Do we need the convenience of just passing an int or should we just require this to be defined explicitly?

linear_operator/operators/kernel_linear_operator.py
```diff
@@ -0,0 +1,385 @@
...
+    :param params: Additional hyperparameters (:math:`\boldsymbol \theta`) or keyword arguments passed into covar_func.
+    """
+
+    def __init__(
+        self,
+        x1: Float[Tensor, "... M D"],
+        x2: Float[Tensor, "... N D"],
+        covar_func: Callable[..., Float[Union[Tensor, LinearOperator], "... M N"]],
+        num_outputs_per_input: Union[int, Tuple[int, int]] = (1, 1),
+        num_nonbatch_dimensions: Dict[str, int] = dict(),
```
Balandat: linters are going to hate creating a dict in the function signature. Suggest using `Optional[Dict[str, int]]= None` instead

linear_operator/operators/kernel_linear_operator.py
```diff
@@ -0,0 +1,385 @@
...
+        param_batch_shapes = dict()
+        param_nonbatch_shapes = dict()
+        for name, val in tensor_params.items():
+            if num_nonbatch_dimensions[name] == 0:
+                param_batch_shapes[name] = val.shape
+                param_nonbatch_shapes[name] = torch.Size([])
+            else:
+                param_batch_shapes[name] = val.shape[: -num_nonbatch_dimensions[name]]
+                param_nonbatch_shapes[name] = val.shape[-num_nonbatch_dimensions[name] :]
+
```
Balandat:
```suggestion
                nonbatch_dim = num_nonbatch_dimensions[name]
                param_batch_shapes[name] = val.shape[: -nonbatch_dim]
                param_nonbatch_shapes[name] = val.shape[-nonbatch_dim :]

```

linear_operator/operators/kernel_linear_operator.py
```diff
@@ -0,0 +1,385 @@
...
+        if len(batch_broadcast_shape):  # Otherwise all tensors are non-batch, and we don't need to expand
+            x1 = x1.expand(*batch_broadcast_shape, *x1.shape[-2:]).contiguous().requires_grad_(x1.requires_grad)
+            x2 = x2.expand(*batch_broadcast_shape, *x2.shape[-2:]).contiguous().requires_grad_(x2.requires_grad)
+            tensor_params = dict(
+                (
+                    name,
+                    val.expand(*batch_broadcast_shape, *param_nonbatch_shapes[name]).requires_grad_(val.requires_grad),
+                )
+                for name, val in tensor_params.items()
+            )
```
Balandat:
```suggestion
            tensor_params = {
                name : val.expand(
                    *batch_broadcast_shape, *param_nonbatch_shapes[name]
                ).requires_grad_(val.requires_grad)
                for name, val in tensor_params.items()
            }
```

linear_operator/operators/kernel_linear_operator.py
```diff
@@ -0,0 +1,385 @@
...
+        self.num_nonbatch_dimensions = num_nonbatch_dimensions
+
+    @cached(name="kernel_diag")
+    def _diagonal(self: Float[LinearOperator, "... M N"]) -> Float[torch.Tensor, "... N"]:
+        # Explicitly compute kernel diag via covar_func when it is needed rather than relying on lazy tensor ops.
+        # We will do this by shoving all of the data into a batch dimension (i.e. compute a ... x N x 1 x 1 kernel)
+        # and then squeeze out the batch dimensions
+        x1 = self.x1.unsqueeze(0).transpose(0, -2)
+        x2 = self.x2.unsqueeze(0).transpose(0, -2)
+        tensor_params = dict((name, val.unsqueeze(0)) for name, val in self.tensor_params.items())
```
Balandat:
```suggestion
        tensor_params = {name: val.unsqueeze(0) for name, val in self.tensor_params.items()}
```

linear_operator/operators/kernel_linear_operator.py
```diff
@@ -0,0 +1,385 @@
...
+
+    @property
+    @cached(name="covar_mat")
+    def covar_mat(self: Float[LinearOperator, "... M N"]) -> Float[Union[Tensor, LinearOperator], "... M N"]:
+        return self.covar_func(self.x1, self.x2, **self.tensor_params, **self.nontensor_params)
+
+    def _get_indices(self, row_index: IndexType, col_index: IndexType, *batch_indices: IndexType) -> torch.Tensor:
+        x1_ = self.x1[(*batch_indices, row_index)].unsqueeze(-2)
+        x2_ = self.x2[(*batch_indices, col_index)].unsqueeze(-2)
+        tensor_params_ = dict((name, val[batch_indices]) for name, val in self.tensor_params.items())
```
Balandat:
```suggestion
        tensor_params_ = {name: val[batch_indices] for name, val in self.tensor_params.items()}
```

linear_operator/operators/kernel_linear_operator.py
```diff
@@ -0,0 +1,385 @@
...
+        # Get the indices of x1 and x2 that matter for the kernel
+        # Call x1[*batch_indices, row_index, :] and x2[*batch_indices, col_index, :]
+        x1 = _x_getitem(self.x1, batch_indices, row_index)
+        x2 = _x_getitem(self.x2, batch_indices, col_index)
+
+        # Call params[*batch_indices, :, :]
+        tensor_params = dict(
+            (name, val[(*batch_indices, *([_noop_index] * self.num_nonbatch_dimensions[name]))])
+            for name, val in self.tensor_params.items()
+        )
```
Balandat:
```suggestion
        tensor_params = {
            name: val[(*batch_indices, *([_noop_index] * self.num_nonbatch_dimensions[name]))]
            for name, val in self.tensor_params.items()
        }
```

linear_operator/operators/kernel_linear_operator.py
```diff
@@ -0,0 +1,385 @@
...
+    ) -> Union[Float[torch.Tensor, "... M C"], Float[torch.Tensor, "... M"]]:
+        return self.covar_mat @ rhs.contiguous()
+
+    def _permute_batch(self, *dims: int) -> LinearOperator:
+        x1 = self.x1.permute(*dims, -2, -1)
+        x2 = self.x2.permute(*dims, -2, -1)
+        tensor_params = dict(
+            (name, val.permute(*dims, *range(-self.num_nonbatch_dimensions[name], 0)))
+            for name, val in self.tensor_params.items()
+        )
```
Balandat:
```suggestion
        tensor_params = {
            name: val.permute(*dims, *range(-self.num_nonbatch_dimensions[name], 0))
            for name, val in self.tensor_params.items()
        }
```

linear_operator/operators/kernel_linear_operator.py
```diff
@@ -0,0 +1,385 @@
...
+            num_outputs_per_input=self.num_outputs_per_input,
+            num_nonbatch_dimensions=self.num_nonbatch_dimensions,
+            **self.tensor_params,
+            **self.nontensor_params,
+        )
+
+    def _unsqueeze_batch(self, dim: int) -> LinearOperator:
+        x1 = self.x1.unsqueeze(dim)
+        x2 = self.x2.unsqueeze(dim)
+        tensor_params = dict((name, val.unsqueeze(dim)) for name, val in self.tensor_params.items())
```
Balandat:
```suggestion
        tensor_params = {name: val.unsqueeze(dim) for name, val in self.tensor_params.items()}
```

linear_operator/operators/linear_operator_representation_tree.py
```diff
@@ -4,11 +4,12 @@
...
     def __init__(self, linear_op):
         self._cls = linear_op.__class__
-        self._kwargs = linear_op._kwargs
+        self._differentiable_kwarg_names = linear_op._differentiable_kwarg_names
+        self._nondifferentiable_kwargs = linear_op._nondifferentiable_kwargs
 
         counter = 0
         self.children = []
-        for arg in linear_op._args:
+        for arg in list(linear_op._args) + list(linear_op._differentiable_kwarg_vals):
```
Balandat:
```suggestion
        for arg in itertools.chain(linear_op._args, linear_op._differentiable_kwarg_vals):
```

Balandat (approved):
I can't say I did check all the indexing logic in exhaustive detail, but hopefully we have some test coverage for that?

linear_operator/operators/kernel_linear_operator.py
gpleiss: We can define it explicitly.

linear_operator/operators/kronecker_product_linear_operator.py
```diff
@@ -62,23 +62,36 @@ def _t_matmul(linear_ops, kp_shape, rhs):
...
+        try:
+            batch_broadcast_shape = torch.broadcast_shapes(*(linear_op.batch_shape for linear_op in linear_ops))
+        except RuntimeError:
+            raise RuntimeError(
+                "Batch shapes of LinearOperators "
+                f"({', '.join([str(tuple(linear_op.shape)) for linear_op in linear_ops])}) "
+                "are incompatible for a Kronecker product."
+            )
+
+        if len(batch_broadcast_shape):  # Otherwise all linear_ops are non-batch, and we don't need to expand
```
Balandat, line 87: This introduced an inconsistency in the type of `linear_ops` (a `list` rather than a `tuple`), which resulted in some downstream breakages in botorch. Fixed in #66.

PR #65, opened 2023-05-31
Block tensor logic

corwinjoy:
Initial work for a block tensor linear operator


corwinjoy:
Sorry about that. Work in progress.

PR #66, opened 2023-06-03
Fix type of KroneckerProductLinearOperator.linear_ops (in some cases)

Balandat:
https://github.com/cornellius-gp/linear_operator/pull/62 introduced an inconsistency of the `linear_ops` property of `KroneckerProductLinearOperator` (by making it a `list` rather than a `tuple` in some cases). This broke some downstream usage of this that relied on this being a tuple.


Balandat:
(could probably have been avoided by better unit test coverage...)

Issue #54, opened 2023-03-26
[Feature Request] Block matrix support


hughsalimbeni:
# 🚀 Feature Request

Represent [TN, TM] tensors by TxT blocks of NxM lazy tensors. While block matrices are supported, the efficient representation is only when there is a diagonal structure over the T dimensions. 

## Motivation

Here is an example that linear_operator cannot deal with: 

```python
import torch
import itertools

T, N, M = 2, 4, 3
As = [torch.rand(N, M) for _ in range(T)]
Bs = [[torch.rand(M, M) for _ in range(T)] for _ in range(T)]
Cs = [torch.rand(N, N) for _ in range(T)]
L = torch.rand(T, T)

A_bl = torch.zeros((N * T, M * T))  # BlockDiag (non-square)
B_bl = torch.zeros((M * T, M * T))  # Dense
C_bl = torch.zeros((N * T, N * T))  # BlockDiag
L_bl = torch.kron(L, torch.eye(N))  # Kroneker

for t in range(T):
    A_bl[N * t : N * (t + 1), M * t : M * (t + 1)] = As[t]
    C_bl[N * t : N * (t + 1), N * t : N * (t + 1)] = Cs[t]

for t1, t2 in itertools.product(range(T), range(T)):
    B_bl[M * t1 : M * (t1 + 1), M * t2 : M * (t2 + 1)] = Bs[t1][t2]

# Desired calculation
print("inefficient method")
print(torch.diag(L_bl @ (C_bl + A_bl @ B_bl @ A_bl.T) @ L_bl.T))
```

This calculation turns up in some multi-output GP models. It has a straightforward efficient implementation:

```python
M_diag = {}
# We only need the diagonal of each block of M
for t1, t2 in itertools.product(range(T), range(T)):
    r = (As[t1].T * (Bs[t1][t2] @ As[t2].T)).sum(0)
    if t1 == t2:
        r += torch.diag(Cs[t1])
    M_diag[(t1, t2)] = r

# The rotation is applied blockwise due to the kron structure
R = {}
for t in range(T):  # we don't need the off-diag blocks
    r = 0
    for i1, i2 in itertools.product(range(T), range(T)):
        r += L[t, i1] * M_diag[(i1, i2)] * L[t, i2]
    R[t] = r

print("fast way")
print(torch.concat([R[t] for t in range(T)]))
```

Currently, this calculation could be implemented inside linear_operator like this

```python
from linear_operator.operators import (
[...]

PR #67, opened 2023-06-05
Block Matrix Linear Operator

corwinjoy:
### Idea
Represent [TN, TM] tensors by TxT blocks of NxM lazy tensors. While block matrices are currently supported, the efficient representation is only when there is a diagonal structure over the T dimensions.

### Pitch
Add a block linear operator class that can keep track of the [T, T] block structure, represented as T^2 lazy tensors of the same shape. Implement matrix multiplication between block matrices as the appropriate linear operators on the blocks.

### Previous Discussion

Issue [https://github.com/cornellius-gp/linear_operator/issues/54](https://github.com/cornellius-gp/linear_operator/issues/54)

### Additional Considerations

In pursuing this, it seems that the base test class checks for many operations beyond what is required to create a LinearOperator. I propose a refactoring of the test class into required / core operations and optional operations. For now, I have created a new core test class CoreLinearOperatorTestCase and have shown what has been excluded by commenting out the relevant code. This idea could also use a review for accuracy.

### DIFF
--- a/linear_operator/operators/block_matrix_linear_operator.py
+++ b/linear_operator/operators/block_matrix_linear_operator.py
@@ -9,11 +9,17 @@ from .dense_linear_operator import DenseLinearOperator
 from .zero_linear_operator import ZeroLinearOperator
 
 
-class MatrixLinearOperator(LinearOperator):
+class BlockMatrixLinearOperator(LinearOperator):
     """
-    A TxT matrix of LinearOperators
+    A TxT block matrix of LinearOperators.
 
-    :param linear_operators: A TxT nested list of linear operators reprsenting a 2-D matrix
+    Idea. Represent [TN, TM] tensors by TxT blocks of NxM lazy tensors.
+
+    Implementation. A block linear operator class that can keep track of the [T, T] block structure,
+    represented as T^2 lazy tensors of the same shape. Implement matrix multiplication between block matrices as
+    the appropriate linear operators on the blocks.
+
+    :param linear_operators: A TxT nested list of linear operators representing a 2-D matrix
     """
 
     def __init__(self, linear_operators: List[List[LinearOperator]]) -> None:
@@ -60,7 +66,7 @@ class MatrixLinearOperator(LinearOperator):
         # A is block [N * T1, M * T2] and B is block [O * S1, P * S2]. If A and B have conformal block counts
         # ie T2==S1 as well as M==O then use the blockwise algorithm. Else use to_dense()
         if isinstance(rhs, self.__class__) and self.num_tasks == rhs.num_tasks and self.block_cols == rhs.block_rows:
-            output = MatrixLinearOperator.create_square_ops_output(T)
+            output = BlockMatrixLinearOperator.create_square_ops_output(T)
             for i in range(T):
                 for j in range(T):
                     out_ij = self.linear_operators[i][0] @ rhs.linear_operators[0][j]
@@ -79,7 +85,7 @@ class MatrixLinearOperator(LinearOperator):
                 P_T = rhs.size(1) // T
                 rhs_blocks_raw = rhs.reshape(T, O_T, T, P_T)
                 rhs_blocks = rhs_blocks_raw.permute(0, 2, 1, 3)
-                rhs_op = MatrixLinearOperator.from_tensor(rhs_blocks, T)
+                rhs_op = BlockMatrixLinearOperator.from_tensor(rhs_blocks, T)
                 return self._matmul(rhs_op)
 
         # Failover implementation. Convert to dense and multiply matricies
@@ -140,7 +146,7 @@ class MatrixLinearOperator(LinearOperator):
             for j in range(self.num_tasks):
                 rows.append(self.linear_operators[j][i].mT)
             out.append(rows)
-        return MatrixLinearOperator(out)
+        return BlockMatrixLinearOperator(out)
 
     def _getitem(self, row_index: IndexType, col_index: IndexType, *batch_indices: IndexType) -> LinearOperator:
         # Perform the __getitem__
@@ -149,7 +155,7 @@ class MatrixLinearOperator(LinearOperator):
         return DenseLinearOperator(res)
 
     @classmethod
-    def from_tensor(cls, tensor: Tensor, num_tasks: int) -> "MatrixLinearOperator":
+    def from_tensor(cls, tensor: Tensor, num_tasks: int) -> "BlockMatrixLinearOperator":
         def tensor_to_linear_op(t: Tensor) -> LinearOperator:
             if torch.count_nonzero(t) > 0:
                 return DenseLinearOperator(t)
--- a/linear_operator/operators/block_matrix_linear_operator.py
+++ b/linear_operator/operators/block_matrix_linear_operator.py
@@ -60,25 +60,30 @@ class BlockMatrixLinearOperator(LinearOperator):
             ops.append(tmp)
         return ops
 
+    def _matmul_two_block_matrix_linear_operators(
+        self: "BlockMatrixLinearOperator",
+        other: "BlockMatrixLinearOperator",
+    ) -> "BlockMatrixLinearOperator":
+        assert self.num_tasks == other.num_tasks
+        assert self.block_cols == other.block_rows
+
+        T = self.num_tasks
+        output = BlockMatrixLinearOperator.create_square_ops_output(T)
+        for i in range(T):
+            for j in range(T):
+                out_ij = self.linear_operators[i][0] @ other.linear_operators[0][j]
+                for k in range(1, T):
+                    out_ij += self.linear_operators[i][k] @ other.linear_operators[k][j]
+                output[i][j] = out_ij
+        return self.__class__(output)
+
     def _matmul(
         self: Float[LinearOperator, "*batch M N"],
         rhs: Union[Float[torch.Tensor, "*batch2 N C"], Float[torch.Tensor, "*batch2 N"]],
     ) -> Union[Float[torch.Tensor, "... M C"], Float[torch.Tensor, "... M"]]:
-
         T = self.num_tasks
 
-        # A is block [N * T1, M * T2] and B is block [O * S1, P * S2]. If A and B have conformal block counts
-        # ie T2==S1 as well as M==O then use the blockwise algorithm. Else use to_dense()
-        if isinstance(rhs, self.__class__) and self.num_tasks == rhs.num_tasks and self.block_cols == rhs.block_rows:
-            output = BlockMatrixLinearOperator.create_square_ops_output(T)
-            for i in range(T):
-                for j in range(T):
-                    out_ij = self.linear_operators[i][0] @ rhs.linear_operators[0][j]
-                    for k in range(1, T):
-                        out_ij += self.linear_operators[i][k] @ rhs.linear_operators[k][j]
-                    output[i][j] = out_ij
-            return self.__class__(output)
-        elif isinstance(rhs, Tensor) and rhs.ndim == 2:
+        if isinstance(rhs, Tensor) and rhs.ndim == 2:
             # Check both matrix dims divisible by T,
             # reshape to (T, T, ), call block multiplication
             if rhs.size(0) % T == 0 and rhs.size(1) % T == 0:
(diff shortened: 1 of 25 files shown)


corwinjoy:
@Balandat After more discussion with @hughsalimbeni and @dannyfriar here is our initial proposal to expand the library as per the previous discussion.
We think it represents a good way to generalize the operator structure to capture nested operations.
We also look forward to any suggestions you may have.
Also tagging: @gpleiss, @jacobrgardner

linear_operator/test/linear_operator_core_test_case.py
```diff
@@ -0,0 +1,518 @@
...
+specifies the tensor(s) needed to define the LinearOperator,
+specifies a _matmul function (how the LinearOperator is applied to a vector),
+specifies a _size function (how big is the LinearOperator if it is represented as a matrix, or batch of matrices), and
+specifies a _transpose_nonbatch function (the adjoint of the LinearOperator).
+(optionally) defines other functions (e.g. logdet, eigh, etc.) to accelerate computations for which efficient
+sturcture-exploiting routines exist.
+
+What follows is a class to test these core LinearOperator operations
+"""
+
```
corwinjoy, line 24: This file is copied from linear_operator_test_case.py. The commented out sections represent non-core operations that I dropped. I think these should be optional operations but I may have misunderstood some tests.

corwinjoy:
Hi, I just wanted to check in and see if you had time to take a look at our proposal and if you had any thoughts?  Thanks again for the feedback so far!
@Balandat @gpleiss

gpleiss:
Sorry for the delay @corwinjoy - this PR is on my todo list for Monday.

linear_operator/operators/block_matrix_linear_operator.py
```diff
@@ -0,0 +1,168 @@
...
+
+    Implementation. A block linear operator class that can keep track of the [T, T] block structure,
+    represented as T^2 lazy tensors of the same shape. Implement matrix multiplication between block matrices as
+    the appropriate linear operators on the blocks.
+
+    :param linear_operators: A TxT nested list of linear operators representing a 2-D matrix
+    """
+
+    def __init__(self, linear_operators: List[List[LinearOperator]]) -> None:
+        assert isinstance(linear_operators, list), f"{self.__class__.__name__} expects a nested list of LinearOperators"
```
gpleiss: Small nit: could we instead assert that `linear_operators` is an iterable?

linear_operator/operators/block_matrix_linear_operator.py
```diff
@@ -0,0 +1,168 @@
...
+        # Check that provided operators all have the same shape
+        T = self.num_tasks
+        for i in range(T):
+            for j in range(T):
+                assert (
+                    linear_operators[i][j].shape[0] == self.block_rows
+                ), "the number of rows much match for all linear operators"
+                assert (
+                    linear_operators[i][j].shape[1] == self.block_cols
+                ), "the number of columns much match for all linear operators"
```
gpleiss: Other small nit: this is a bit of checking that's happening under the hood, which might get a bit expensive. Would it be possible to do this checking only if we're in debug mode (using the context manager `with linear_operator.settings.debug(True)`?

If that's undesirable I'm happy to keep it as is, but just seeing if there's ways to lower the python overhead.

linear_operator/operators/block_matrix_linear_operator.py
```diff
@@ -0,0 +1,168 @@
...
+                O_T = rhs.size(0) // T
+                P_T = rhs.size(1) // T
+                rhs_blocks_raw = rhs.reshape(T, O_T, T, P_T)
+                rhs_blocks = rhs_blocks_raw.permute(0, 2, 1, 3)
+                rhs_op = BlockMatrixLinearOperator.from_tensor(rhs_blocks, T)
+                return self._matmul(rhs_op)
+
+        # Failover implementation. Convert to dense and multiply matricies
+        A = self.to_dense()
+        B = rhs.to_dense()
```
gpleiss: `rhs` is a tensor, so it shouldn't have a `to_dense` method?

linear_operator/operators/block_matrix_linear_operator.py
```diff
@@ -0,0 +1,168 @@
...
+                # If O and P are both divisible by T,
+                # then interpret B as a [O//T * T, P//T * T] block matrix
+                O_T = rhs.size(0) // T
+                P_T = rhs.size(1) // T
+                rhs_blocks_raw = rhs.reshape(T, O_T, T, P_T)
+                rhs_blocks = rhs_blocks_raw.permute(0, 2, 1, 3)
+                rhs_op = BlockMatrixLinearOperator.from_tensor(rhs_blocks, T)
+                return self._matmul(rhs_op)
+
+        # Failover implementation. Convert to dense and multiply matricies
```
gpleiss, line 91: I imagine we'll hit this failover implementation any time we have batched linear operators/rhs (as we are explicitly checking for `ndims=2`). Ideally, the implementation would someday work for the batched case as well. In the meantime, it's probably good to throw an error to alert the user that we are densely instantiating the linear operator.

linear_operator/operators/block_matrix_linear_operator.py
gpleiss: Oh I see... rhs does get converted into a linear operator. (The type hit should probably be updated)
This will probably fail when `rhs` is a tensor but `rhs.dim() > 2`. It may be wise to do some type checking here.

linear_operator/operators/block_matrix_linear_operator.py
```diff
@@ -0,0 +1,168 @@
...
+        A = self.to_dense()
+        B = rhs.to_dense()
+        res = A @ B
+        return res
+
+    def matmul(
+        self: Float[LinearOperator, "*batch M N"],
+        other: Union[Float[Tensor, "*batch2 N P"], Float[Tensor, "*batch2 N"], Float[LinearOperator, "*batch2 N P"]],
+    ) -> Union[Float[Tensor, "... M P"], Float[Tensor, "... M"], Float[LinearOperator, "... M P"]]:
+        return self._matmul(other)
```
gpleiss, line 116: Is this necessary?

linear_operator/test/linear_operator_core_test_case.py
gpleiss: I'm a fan of this! Could you refactor the current LO test case to inherit from the core case?

corwinjoy: Will do!

linear_operator/operators/block_matrix_linear_operator.py
corwinjoy: Done!

linear_operator/operators/block_matrix_linear_operator.py
corwinjoy: Makes sense. Thanks for the suggestion!

linear_operator/operators/block_matrix_linear_operator.py
corwinjoy: OK. I have updated the type signature. (But only in the comments since this has to conform to the base class). I haven't added explicit type-checking here since I expect the matrix multiplication of A@B to give an error message if the dimension sizes don't line up. (I think that explicit checking may just get unnecessarily messy here).

linear_operator/operators/block_matrix_linear_operator.py
corwinjoy: I have put an assert check into the failover case for now. There may be combinations of this new block operator + regular linear operator that we want to allow (inefficiently) for now. 
@dannyfriar - I think that makes sense.

linear_operator/operators/block_matrix_linear_operator.py
corwinjoy: Unfortunately, I think it is. The base method wants to perform a matmul via broadcasting and a representation tree which this operator doesn't support. I tried to get this operator to have a representation tree but I couldn't figure out how to do it in a general way as needed  by this class.

linear_operator/operators/block_matrix_linear_operator.py
corwinjoy: Thanks for the offline feedback Danny. Will leave this as an assert for now.

corwinjoy:
@gpleiss Thanks for all the great feedback! I believe I have addressed all these and would appreciate a second look when you have time.

linear_operator/operators/block_matrix_linear_operator.py
gpleiss: Hmm alright... could you add a comment in the code that mentions this? (I want to be sure that we don't accidentally delete this function down the road!)

gpleiss: Actually, I can do it and then we're ready to merge :)

linear_operator/operators/block_matrix_linear_operator.py
```diff
@@ -0,0 +1,176 @@
...
+        assert B.ndim <= 2
+
+        res = A @ B
+        return res
+
+    def matmul(
+        self: Float[LinearOperator, "*batch M N"],
+        other: Union[Float[Tensor, "*batch2 N P"], Float[Tensor, "*batch2 N"], Float[LinearOperator, "*batch2 N P"]],
+    ) -> Union[Float[Tensor, "... M P"], Float[Tensor, "... M"], Float[LinearOperator, "... M P"]]:
+        return self._matmul(other)
```
gpleiss, line 116:
```suggestion
        # The base method wants to perform a matmul via broadcasting and a 
        # representation tree which this operator doesn't support.
        return self._matmul(other)
```

gpleiss (approved):

gpleiss:
@corwinjoy after playing around with this PR some more, and dealing with the typeguard errors, I made 2 commits that change around some of the internals. The first is hopefully not to controversial, the second one might be :)

1. **Rearranging the logic of _matmul.** The purpose of the `_matmul` function is defining a matrix-vector product that can be used by iterative algorithms (e.g. CG), so - upon further reflection - I think that `_matmul` should only accept Tensor rhs, and it should also output a Tensor as well.

I rearranged some of the logic, in what I hope won't affect anything that you've already written.
- The previous base case of `_matmul` now lives in a private function called `_matmul_two_block_matrix_linear_operators`, which takes in two BlockMatrixLOs and outputs a BlockMatrixLO
- The public `matmul` function now has logic to call this new private function if `other` is an appropriate BlockMatrixLO. Otherwise it calls the `_matmul` function.

2. **Changing the constructor.** (slightly more controversial). There were some typing issues with the constructor, since it was taking in a list of lists of linear operators. The arguments to all other LinearOperators are expected to be linear operators or Tensors. (I noticed that you had to overwrite the `representation` function to accommodate this). There are unfortunately a lot of (necessary) hacks throughout LinearOperator that do expect the arguments to be LinearOperators and Tensors, and I'm a bit nervous that there could be some unintended bugs by not adhering to this pattern. (Historically, these bugs emerge as gradient issues or type-casting issues.)

My proposed solution: the user passes in a flattened list of `T^2` linear operators into the constructor. It's a little more unintuitive for the user (passing in a list of lists is more natural, I agree), but it actually made some of the code quite a bit simpler.


Anyways, both commits are now part of this branch. Let me know your thoughts, and we can always revert and figure out different solutions.

corwinjoy:
@gpleiss Thanks for the detailed review l. It all sounds good except for possibly the constructor which I need to think about. I am traveling right now but should be able to take a detailed look by Wednesday. @dannyfriar any thoughts? @hughsalimbeni

dannyfriar:
> @gpleiss Thanks for the detailed review l. It all sounds good except for possibly the constructor which I need to think about. I am traveling right now but should be able to take a detailed look by Wednesday. @dannyfriar any thoughts? @hughsalimbeni

I also think that passing in a list of lists is more natural. One thing we'd like to do with this is have the ability to rotate the blocks - this is simpler with the list of lists but I'm sure it can be made to work with a flat list too. @gpleiss out of curiosity what are the hacks that you mentioned that require LOs/tensors?

corwinjoy:
@gpleiss Thanks for the code changes!  Looking at these, they make sense. I like the better matmul function. The flattened list of operators for the constructor is not my first choice, but if that is what needs to happen for compatibility I can live with it. We do have a from_tensor helper constructor already and if we can add a nested list helper later if we find out we need it. Like @dannyfriar I am a bit curious about the hacks that require this format if you do have time to explain. At any rate, I am happy with the changes as they are.

CarloGraziani:
This looks pretty close to something that I also need: a block-Toeplitz linear operator, which arises in multi-output GPs with kernels that are stationary over their 1-D input space. This is relevant to LLM work, where the input space is sequence location and the output space is token categorical probability.  

If I were to simply replicate LazyTensor blocks as required in the list of N**2 blocks to make a block-Toeplitz structure, would that work, or would it break something (like autograd treating the blocks as independent variables when they are not)?

PR #70, opened 2023-06-08
Add short path for LinearOperator.repeat()

Turakar:
This adds a short path for LinearOperator.repeat() if we would otherwise create a new `BatchRepeatLinearOperator` which would be a no-op.

`repeat()` is called like this for example by the constructor of `KroneckerProductLinearOperator`.

By reducing one level of indirection, we gain two benefits: a) slightly better performance (obviously), and b) some more optimizations by calling native functions instead of default `LinearOperator` functions, e.g. for `_diagonal()`.

### DIFF
--- a/linear_operator/operators/_linear_operator.py
+++ b/linear_operator/operators/_linear_operator.py
@@ -2023,15 +2023,15 @@ class LinearOperator(object):
         """
         from .batch_repeat_linear_operator import BatchRepeatLinearOperator
 
+        # Short path if no repetition is necessary
+        if all(x == 1 for x in sizes) and len(sizes) == self.dim():
+            return self
+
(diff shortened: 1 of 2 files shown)


Balandat (commented):
This makes sense to me. Should we add a unit test for this to check that we indeed return the same object rather than a `BatchRepeatLinearOperator` in this case?

gpleiss (changes requested):
@Turakar please add a unit test like @Balandat suggests, and then we'll merge!

Turakar:
Seems like there were no unit tests before, so I added tests for both cases: A No-Op case and a true case. I also allowed the `repeat(1, 1)` case as this is a no-op, too.

Balandat (approved):

PR #71, opened 2023-07-14
Fix normalization of initial guess in linear_cg()

Turakar:
Seems like while `rhs` was normalized, `initial_guess` was not. As such, initializing `linear_cg()` with a good initial guess (e.g. the solution) was of no use. This PR also normalizes `initial_guess` and adds a corresponding test.

### DIFF
--- a/test/utils/test_linear_cg.py
+++ b/test/utils/test_linear_cg.py
@@ -143,6 +143,26 @@ class TestLinearCG(unittest.TestCase):
                 approx_eigs = torch.linalg.eigvalsh(t_mats[j, i])
                 self.assertTrue(torch.allclose(eigs, approx_eigs, atol=1e-3, rtol=1e-4))
 
+    def test_batch_cg_init(self):
+        batch = 5
+        size = 100
+        matrix = torch.randn(batch, size, size, dtype=torch.float64)
+        matrix = matrix.matmul(matrix.mT)
+        matrix.div_(matrix.norm())
+        matrix.add_(torch.eye(matrix.size(-1), dtype=torch.float64).mul_(1e-1))
+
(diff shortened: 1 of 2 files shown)


Turakar:
On a side note: For my example, initializing `linear_cg()` with the solve from the previous iteration of Gaussian Process training with `skip_logdet_forward(True)` was of no use regarding convergence speed.

test/utils/test_linear_cg.py
```diff
@@ -143,6 +143,26 @@ def test_batch_cg_with_tridiag(self):
...
+    def test_batch_cg_init(self):
+        batch = 5
+        size = 100
+        matrix = torch.randn(batch, size, size, dtype=torch.float64)
+        matrix = matrix.matmul(matrix.mT)
+        matrix.div_(matrix.norm())
+        matrix.add_(torch.eye(matrix.size(-1), dtype=torch.float64).mul_(1e-1))
+
+        # Initial solve
+        rhs = torch.randn(batch, size, 50, dtype=torch.float64)
```
Balandat, line 155: The randomness here seems like it could easily result in a flaky test - have you run this in a stress test (e.g. a few hundred iterations of this) to check this doesn't happen? Alternatively you could fix the seed.

Balandat (approved):
This makes sense to me

test/utils/test_linear_cg.py
Turakar: Seems to me like the seed is already fixed: https://github.com/Turakar/linear_operator/blob/9b30670d55740499405c80ddeb9cc66fd36fd7df/test/utils/test_linear_cg.py#L18

PR #72, opened 2023-07-20
Fix _t_matmul() in MaskedLinearOperator and add test

Turakar:
I am sorry, but it seems like #69 contained a bug.


Balandat (approved):
lgtm

PR #74, opened 2023-08-11
Fix type error when converting interpolated_linear_operator to double precision

CY-Zhang:
- Fix type error when converting interpolated_linear_operator to double precision, as discussed in https://github.com/cornellius-gp/gpytorch/issues/1958
- Added to() method in interpolated_linear_operator.py to overwrite the default to() method in _linear_operator. For each arg, the dtype will be converted only when the original dtype and target dtype are both int/float, to avoid converting integer index matrices to float.
- Added new unit test case under linear_operator_test_case.py to test whether after calling to(torch.float64), the linear operator is still functional and can be converted to numpy array. Raise TypeError if the test failed.

### DIFF
--- a/linear_operator/operators/interpolated_linear_operator.py
+++ b/linear_operator/operators/interpolated_linear_operator.py
@@ -10,6 +10,7 @@ from torch import Tensor
 
 from ..utils import sparse
 from ..utils.broadcasting import _pad_with_singletons
+from ..utils.generic import _to_helper
 from ..utils.getitem import _noop_index
 from ..utils.interpolation import left_interp, left_t_interp
 from ._linear_operator import IndexType, LinearOperator
@@ -454,3 +455,28 @@ class InterpolatedLinearOperator(LinearOperator):
         res = left_interp(self.left_interp_indices, self.left_interp_values, base_samples).contiguous()
         batch_iter = tuple(range(res.dim() - 1))
         return res.permute(-1, *batch_iter).contiguous()
+
+    def to(self: Float[LinearOperator, "*batch M N"], *args, **kwargs) -> Float[LinearOperator, "*batch M N"]:
+
+        # Overwrite the to() method in _linear_operator to avoid converting index matrices to float.
+        # Will only convert both dtype and device when arg and dtype are both int/float.
+        # Otherwise, will only convert device.
+
+        device, dtype = _to_helper(*args, **kwargs)
+
+        new_args = []
+        new_kwargs = {}
+        for arg in self._args:
+            if hasattr(arg, "to"):
+                if arg.dtype.is_floating_point == dtype.is_floating_point:
+                    new_args.append(arg.to(dtype=dtype, device=device))
+                else:
+                    new_args.append(arg.to(device=device))
+            else:
+                new_args.append(arg)
+        for name, val in self._kwargs.items():
+            if hasattr(val, "to"):
+                new_kwargs[name] = val.to(dtype=dtype, device=device)
+            else:
+                new_kwargs[name] = val
+        return self.__class__(*new_args, **new_kwargs)
--- a/linear_operator/operators/interpolated_linear_operator.py
+++ b/linear_operator/operators/interpolated_linear_operator.py
@@ -468,7 +468,7 @@ class InterpolatedLinearOperator(LinearOperator):
         new_kwargs = {}
         for arg in self._args:
             if hasattr(arg, "to"):
-                if arg.dtype.is_floating_point == dtype.is_floating_point:
+                if hasattr(arg, "dtype") and arg.dtype.is_floating_point == dtype.is_floating_point:
                     new_args.append(arg.to(dtype=dtype, device=device))
                 else:
                     new_args.append(arg.to(device=device))
--- a/linear_operator/test/linear_operator_test_case.py
+++ b/linear_operator/test/linear_operator_test_case.py
@@ -1255,6 +1255,5 @@ class LinearOperatorTestCase(RectangularLinearOperatorTestCase):
             linear_op = linear_op.to(torch.float64)
             linear_op.numpy()
         except RuntimeError:
-            raise TypeError(f"Could not convert {type(linear_op)} to double.")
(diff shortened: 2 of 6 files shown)


linear_operator/operators/interpolated_linear_operator.py
```diff
@@ -454,3 +455,28 @@ def zero_mean_mvn_samples(
...
+        # Will only convert both dtype and device when arg and dtype are both int/float.
+        # Otherwise, will only convert device.
+
+        device, dtype = _to_helper(*args, **kwargs)
+
+        new_args = []
+        new_kwargs = {}
+        for arg in self._args:
+            if hasattr(arg, "to"):
+                if arg.dtype.is_floating_point == dtype.is_floating_point:
```
Balandat: It could be that `arg` has a `to` attribute but doesn't have a `dtype` attribute (e.g. if it is some other non-tensor or non-module object that just happens to have a `to` method). Not sure exactly how we'd handle this but we probably should? Maybe we want to catch and pass on exceptions here?

linear_operator/test/linear_operator_test_case.py
```diff
@@ -1247,3 +1247,12 @@ def test_svd(self):
             for arg, arg_copy in zip(linear_op.representation(), linear_op_copy.representation()):
                 if arg_copy.requires_grad and arg_copy.is_leaf and arg_copy.grad is not None:
                     self.assertAllClose(arg.grad, arg_copy.grad, **self.tolerances["svd"])
+
+    def test_to_double(self):
+        # test if the linear_op is still functional after convert to double
+        linear_op = self.create_linear_op()
+        try:
+            linear_op = linear_op.to(torch.float64)
+            linear_op.numpy()
```
Balandat, line 1256: Should we also test that the dtype was properly updated? 

Also, we probably want to overwrite this for the interpolated operator to verify that the index tensors were not moved to floating point type

Balandat (commented):
Thanks for putting this together!

linear_operator/operators/interpolated_linear_operator.py
CY-Zhang: That's a good point. Maybe I can add another condition to first check if the `arg` has a `dtype` attribute, before the `is_floating_point` check? If the `arg` does not have a `dtype` attribute, we directly jump to the `device` conversion.

linear_operator/test/linear_operator_test_case.py
CY-Zhang: Make sense, I will include test to check if dtype is updatedm, and check if index tensors are still integers for the interpolated operator case.

linear_operator/test/linear_operator_test_case.py
```diff
@@ -1247,3 +1247,14 @@ def test_svd(self):
...
                     self.assertAllClose(arg.grad, arg_copy.grad, **self.tolerances["svd"])
+
+    def test_to_double(self):
+        # test if the linear_op is still functional and converted to torch.float64 after calling to(torch.float64).
+        linear_op = self.create_linear_op()
+        try:
+            linear_op = linear_op.to(torch.float64)
+            linear_op.numpy()
+        except RuntimeError:
+            raise TypeError(f"Could not convert {type(linear_op)} to double.")
```
Balandat: This seems more like a `RuntimeError` than like a `TypeError`
```suggestion
            raise RuntimeError(f"Could not convert {type(linear_op)} to double.")
```

test/operators/test_interpolated_linear_operator.py
```diff
@@ -48,6 +48,23 @@ def evaluate_linear_op(self, linear_op):
...
 
+    def test_to_double(self):
+        # overwrite the test_to_double under LinearOperatorTestCase.
+        # specifically check if the index matrices are still integer after conversion.
+        linear_op = self.create_linear_op()
+        try:
+            linear_op = linear_op.to(torch.float64)
+            linear_op.numpy()
+        except RuntimeError:
+            raise TypeError(f"Could not convert {type(linear_op)} to double.")
```
Balandat:
```suggestion
            raise RuntimeError(f"Could not convert {type(linear_op)} to double.")
```

test/operators/test_interpolated_linear_operator.py
```diff
@@ -48,6 +48,23 @@ def evaluate_linear_op(self, linear_op):
...
+        if linear_op.dtype != torch.float64:
+            raise TypeError(f"Could not convert {type(linear_op)} to double.")
+        if (
+            linear_op.left_interp_indices.dtype == torch.float64
+            or linear_op.right_interp_indices.dtype == torch.float64
+        ):
+            raise TypeError(f"Index tensor in {type(linear_op)} converted to double.")
```
Balandat: You probably want to use sth like this here to be consistent with the general unit test setup
```suggestion
        self.assertFalse(linear_op.left_interp_indices.dtype.is_floating_point)
        self.assertFalse(linear_op.right_interp_indices.dtype.is_floating_point)
```

linear_operator/test/linear_operator_test_case.py
```diff
@@ -1247,3 +1247,14 @@ def test_svd(self):
...
+        if linear_op.dtype != torch.float64:
+            raise TypeError(f"Could not convert {type(linear_op)} to double.")
```
Balandat:
```suggestion
        self.assertEqual(linear_op.dtype, torch.float64)
```

Balandat (commented):
Looks like there are some test failures; once those are fixed this is ready to go in.

CY-Zhang:
> Looks like there are some test failures; once those are fixed this is ready to go in.

Yeah, it seems besides the interpolated_linear_operator, the dtype conversion method will also cause some error when being applied to masked_linear_operator (update boolean mask matrices to double) and identity_linear_operator (dtype saved in kwargs and did not get updated). I will do similar fix as the interpolated_linear_operator on these two linear_operators.

Balandat:
Thanks for tying up the loose ends here

Balandat (approved):

PR #97, opened 2024-07-11
Typeguard 4.3.0

### LINKED ISSUE #84: Upgrade required typeguard version
Greetings, 

LinearOperator seems to require `typeguard~=2.13.3`, which is from December 2021 and typeguard is currently at v4.1.5. It would be nice to upgrade this dependency to something more recent. 

Thank you.

ChristianWirthAumovio:
Fixes #84 by updating typeguard to 4.X

This is important because the typeguard requirement is in conflict with ydata-profiling: https://github.com/ydataai/ydata-profiling/blob/cc7b9b36364fdcb6db6a23db5d83ad3c6e2e3c8f/requirements.txt#L23 - a commonly used EDA package.
This, in turn, prevents updates to pandas and numpy.

All unittests passed

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -40,7 +40,6 @@ except ImportError:
 install_requires += [
     "scipy",
     "jaxtyping==0.2.19",
-    "typeguard~=4.3.0",
     "mpmath>=0.19,<=1.3",  # avoid incompatibiltiy with torch+sympy with mpmath 1.4
 ]
 
--- a/setup.py
+++ b/setup.py
@@ -90,7 +90,13 @@ setup(
             "sphinx-autodoc-typehints",
             "uncompyle6<=3.9.0",
         ],
-        "test": ["flake8==5.0.4", "flake8-print==5.0.0", "pytest"],
+        "test": [
+            "flake8==5.0.4",
+            "flake8-print==5.0.0",
+            "pytest",
+            "typeguard~=2.13.3"  # jaxtyping seems to only be compatible with older typeguard versions
+            # https://github.com/patrick-kidger/jaxtyping/commit/77c263c3def8ea3bcb7d7642c5a8402c16cf76fb
+        ],
     },
     test_suite="test",
 )
--- a/setup.py
+++ b/setup.py
@@ -40,7 +40,7 @@ except ImportError:
 install_requires += [
     "scipy",
     "jaxtyping>=0.2.9",
-    "typeguard~=2.13.3",
+    "typeguard~=4.3.0",
     "mpmath>=0.19,<=1.3",  # avoid incompatibiltiy with torch+sympy with mpmath 1.4
 ]
 


setup.py
```diff
@@ -40,7 +40,7 @@
 install_requires += [
     "scipy",
     "jaxtyping==0.2.19",
-    "typeguard~=2.13.3",
+    "typeguard~=4.3.0",
```
tmct: I'm all for raising the upper bound, but I'm not sure of the benefit of raising the lower bound so high, especially if no other changes were needed - that typeguard version only arrived in May 2024.

I wonder if the right balance of permissibility may look more like `>=2.13.3,<5` or even `>=3,<5`?

(I'm not a maintainer, sorry...)

tmct:
Hang on a minute, this is only used in tests? We can move this from install_requires to being a dependency of only the test environment!

And - it looks like `>=3.1,<5` might be a nice balance, given that the commented out code [here](https://github.com/cornellius-gp/linear_operator/blob/4a55f467fc3f2240e233b8e5db8638866981fedb/test/operators/test_kronecker_product_added_diag_linear_operator.py#L19) wants to use that minimum

tmct:
Probably easier to describe that by making the change - what do you think? #98 

(And if it's a test requirement, we don't have to over think the permissiveness of the lower bound!)

ChristianWirthAumovio:
> Probably easier to describe that by making the change - what do you think? #98
> 
> (And if it's a test requirement, we don't have to over think the permissiveness of the lower bound!)

I think your solution is good, as this fully resolves the conflict also for the future (as long as one is not doing testing).
However, i'm fine with both PRs and just want to see the problem resolved 😄

ChristianWirthAumovio:
Moving forward with this or or #98 would be of great help, as it blocks some sever downstream issues. Anything preventing this from the next steps? (@gpleiss ?)

tmct:
We are also eager to see either change made, please. (In the meantime we have been using `uv`'s [dependency overrides](https://github.com/astral-sh/uv?tab=readme-ov-file#dependency-overrides) to create functioning environments.)

gpleiss:
@ChristianWirthContinental @tmct the locked version of typeguard was intentional (but poorly documented in the code). Newer versions of typeguard aren't compatible with jaxtyping. See:

- https://github.com/patrick-kidger/jaxtyping/pull/175/commits/ff3ded32885d1f6d41d796bb56ec4db656a04ee5
- https://github.com/patrick-kidger/jaxtyping/commit/77c263c3def8ea3bcb7d7642c5a8402c16cf76fb

It's worth noting that typeguard versioning has caused [previous issues](https://github.com/cornellius-gp/linear_operator/pull/60#issuecomment-1533245504). As a workaround, we can require the specific locked version of typeguard in our testing code (where we need the specific version for jaxtyped tests) but we don't have to lock down typeguard for the main package dependencies (where jaxtyping is strictly used for annotations only).

gpleiss (approved):

tmct:
Thank you!

n.b. if you ever upgrade your min version of jaxtyping beyond I think 0.2.22, as things stand I think we'd hit the same problem again through transitive dependencies being tightened... So hopefully jaxtyping can be made compatible before then!

ChristianWirthAumovio:
@gpleiss Thank you very much! If you don't mind releasing this, i would try to get the new version into botorch to resolve the version conflicts in the dependency chain.

gpleiss:
Done

PR #98, opened 2024-07-18
Move typeguard to being a test requirement only

### LINKED ISSUE #84: Upgrade required typeguard version

tmct:
Fixes #84 

It doesn't seem like this is needed in the library proper? In which let's just take it out of install_requires.


gpleiss:
Duplicate of #97

PR #105, opened 2025-01-22
Unpin jaxtyping

saitcakmak:
The required jaxtyping version conflicts with the requirements of other libraries. The code appears to be compatible with newer versions, so attempting to unpin to see if anything breaks.

### DIFF
--- a/.github/workflows/run_type_checked_test_suite.yml
+++ b/.github/workflows/run_type_checked_test_suite.yml
@@ -21,7 +21,7 @@ jobs:
     - name: Set up Python
       uses: actions/setup-python@v2
       with:
-        python-version: "3.8"
+        python-version: "3.10"
     - name: Install dependencies
       run: |
         if [[ ${{ matrix.pytorch-version }} = "latest" ]]; then
@@ -32,4 +32,4 @@ jobs:
         pip install -e ".[test]"
     - name: Run unit tests
       run: |
-        pytest ${{ inputs.files_to_test }} --jaxtyping-packages=linear_operator,typeguard.typechecked
+        pytest ${{ inputs.files_to_test }} --jaxtyping-packages=typeguard.typechecked
--- a/.github/workflows/run_type_checked_test_suite.yml
+++ b/.github/workflows/run_type_checked_test_suite.yml
@@ -25,9 +25,10 @@ jobs:
     - name: Install dependencies
       run: |
         if [[ ${{ matrix.pytorch-version }} = "latest" ]]; then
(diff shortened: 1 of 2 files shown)


gpleiss:
Hmm I'm not sure why the docs build is failing. Otherwise lgtm

saitcakmak:
Actually, even after unpin, the CI seems to be running with 0.2.19 because that's the latest version available in Py3.8, which is what the CI uses. So, the CI passing here is meaningless. I'm getting some weird `undefined value torch` errors when trying to run `pytest --jaxtyping-packages=linear_operator,typeguard.typechecked` locally with Py3.10 & jaxtyping 0.2.36

cc @Balandat

saitcakmak:
Removing `linear_operator` from the command, `pytest --jaxtyping-packages=typeguard.typechecked` does pass (as in has the same 3 failures as before) with Py3.10 & jaxtyping 0.2.36. I guess we need to update the CI along with the pin

saitcakmak:
Ok, CI is also happy (except for docs, which seems unrelated) and running with latest jaxtyping now

.github/workflows/run_type_checked_test_suite.yml
```diff
@@ -21,15 +21,16 @@ jobs:
...
+        python-version: "3.10"
     - name: Install dependencies
       run: |
         if [[ ${{ matrix.pytorch-version }} = "latest" ]]; then
-          pip install --pre torch -f https://download.pytorch.org/whl/nightly/cpu/torch_nightly.html;
+          pip install --pre torch -f https://download.pytorch.org/whl/nightly/cpu/torch_nightly.html
         else
-          pip install torch==1.11+cpu -f https://download.pytorch.org/whl/torch_stable.html;
+          pip install torch==1.11+cpu -f https://download.pytorch.org/whl/torch_stable.html
+          pip install "numpy<2"
```
Balandat, line 31: why is numpy < 2 needed here?

saitcakmak: torch=1.11 is not compatible with numpy>2 but it does not explicitly require <2 either, so we end up with CI failing with `RuntimeError: Numpy is not available` on `tensor.numpy()` calls despite `numpy==2.2.2` being installed. See https://github.com/cornellius-gp/linear_operator/actions/runs/12961267021/job/36156321493

Balandat: I see. torch 1.11 is rather ancient but I guess this is only relevant for the CI testing, so this is fine by me.

Balandat (approved):

PR #109, opened 2025-01-28
Fix docs build

saitcakmak:
Updates sphinx configuration to:
- Make jaxtyping parsing compatible with the latest version.
- Update the link to pytorch artifacts.
- Suppress a couple types of warnings

Also removes unused uncompyle6 dependency from the setup.py.


Balandat (approved):
thanks!

PR #117, opened 2025-11-26
increase the default max Cholesky size to 50000

### LINKED ISSUE #116: [Inference] The Default Threshold for Cholesky Decomposition
The default threshold for Cholesky decomposition (as opposed to CG) is 800, which seems to be very small.

https://github.com/cornellius-gp/linear_operator/blob/dbed373575224988a5da4fe9a67c44ae153849f0/linear_operator/settings.py#L392-L400

I was a bit surprised that an exact GP trained on 1000 data points was using CG for inference. Because I somehow vaguely remember this threshold was increased to `6000` a long time ago. (I tried to find relevant issues/PRs for reference but I couldn't. So maybe it's my hallucination.)

In any case, this threshold was set 6 years ago. Hardware are much better nowadays. Any thoughts on increasing this threshold? Probably need to benchmark inference speed again.

kayween:
as titled

### DIFF
--- a/linear_operator/settings.py
+++ b/linear_operator/settings.py
@@ -331,6 +331,7 @@ class fast_computations:
     .. _GPyTorch Blackbox Matrix-Matrix Gaussian Process Inference with GPU Acceleration:
         https://arxiv.org/pdf/1809.11165.pdf
     """
+
     covar_root_decomposition = _fast_covar_root_decomposition
     log_prob = _fast_log_prob
     solves = _fast_solves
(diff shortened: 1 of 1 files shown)


gpleiss:
50000 seems like too much for older GPUs. Is there some way we could try 50000, and then adjust if we get OOM errors?

kayween:
Sounds reasonable. I will put together a prototype and update this PR.

Balandat:
Also, lots of people run stuff on CPU and so we definitely want to be careful about cranking the size up too much by default...

Issue #113, opened 2025-10-08
[Proposal] Remove jaxtyping dependency


Balandat:
`jaxtyping` was added as a dependency in https://github.com/cornellius-gp/linear_operator/pull/42. Since then I have not seen any updates that uses that functionality. But it caused some maintenance work both for this repo (https://github.com/search?q=repo%3Acornellius-gp%2Flinear_operator+jaxtyping&type=commits) and has also recently led to a breakage in our internal codebase b/c of some incompatibilities between jaxtypig and jax due to some numpy version issue. I have a strong sense that the added complexity and maintenance headache is not worth the trouble. 

Proposal: Remove the jaxtyping dependency. 

Options: 
- Straightforward: Remove typing code and dependency outright
- Workaround: Manually create no-op `Float`/`Long` type classes to keep the jaxtyping syntax. 

@gpleiss, @corwinjoy, any thoughts?

Balandat:
@gpleiss, @corwinjoy - speak now or forever hold your peace.

corwinjoy:
Removing this sounds reasonable to me. When I added these checks I was hoping for much stronger typing on the shapes and sizes. Unfortunately, many of the routines have quite variable output which blocks strong type checks. Also, the errors are not that easy to read. So, if you feel it is better removed, I understand.

PR #121, opened 2026-02-03
Remove jaxtyping dependency

Balandat:
Gets rid of the jaxtyping dependency as proposed in #113 (due to limited value and causing dependency issues in some places).


Balandat:
`run_small_type_checked_test_suite` shows up as expected but of course won't complete since removed

.gitignore
```diff
@@ -1,3 +1,6 @@
+# Claude
+CLAUDE.md
+
```
kayween, line 3: I am assuming this is the prompt for Claude? Let's remove it.

jacobrgardner (approved):
This was a pretty big one of very "samey" changes but I scrolled through slowly and they all looked right to me.

kayween:
Looks good to me as well.

.gitignore
Balandat: I guess I'm adding it here so the prompt doesn't get included when people use claude, do you want me to remove it from .gitignore so we can consider committing a version of the prompt?

Balandat: Let me merge this for now; we can remove CLAUDE.md from .gitignore in a separate PR if we think that's the right move.

kayween: I don't think we should commit the prompt.

I initially thought these lines were added by mistake (since they seemed to be unrelated to this PR), and so I wanted to flagged these lines. But it looks like this gitignore entry was added intentionally. So we are all set here!

PR #122, opened 2026-02-03
Speed up MatmulLinearOperator.to_dense() for DiagLazyTensors

Balandat:
This can help substantially if some of the operators involved are `DiagLinearOperator`s.
This fixes the issue reported in https://github.com/meta-pytorch/botorch/issues/3159


Balandat:
Only the last commit is relevant here; the rest comes from https://github.com/cornellius-gp/linear_operator/pull/121

jacobrgardner:
Can we rebase this once the jaxtyping removal goes in? This is a little crazystyle to review at the moment since it touches basically every file in the repo.

Edit: Oh I see the point of your comment now, we can just review the diff from the last commit.

jacobrgardner (approved):
Yeah approved assuming the last commit is the only real diff here, that's an obvious speed win.

PR #124, opened 2026-02-17
Make linear operators pickleable

### LINKED ISSUE #2717: [Bug] Can't pickle function CholLinearOperator.to_dense
# 🐛 Bug

Pickling of variational strategy fails since v1.15.0.

## To reproduce

The skorch CI for the gpytorch integration started failing recently and I could narrow it down to the v1.15.0 release of gpytorch. To reproduce:

```python
import pickle

import numpy as np
import torch
import gpytorch
from skorch.probabilistic import GPRegressor

class VariationalRegressionModule(gpytorch.models.ApproximateGP):
    """GP regression for variational inference"""
    def __init__(self, inducing_points, eps=1e-6):
        variational_distribution = gpytorch.variational.CholeskyVariationalDistribution(
            inducing_points.size(0))
        variational_strategy = gpytorch.variational.VariationalStrategy(
            self,
            inducing_points,
            variational_distribution,
            learn_inducing_locations=True,
        )
        super().__init__(variational_strategy)
        self.mean_module = gpytorch.means.ConstantMean()
        self.covar_module = gpytorch.kernels.ScaleKernel(
            gpytorch.kernels.RBFKernel(eps=eps))

    def forward(self, x):
        mean_x = self.mean_module(x)
        covar_x = self.covar_module(x)
        return gpytorch.distributions.MultivariateNormal(mean_x, covar_x)

n_samples = 60
X = np.linspace(-8, 8.01, n_samples).astype(np.float32)
y = (np.sin(X) + np.random.randn(len(X)) * 0.2).astype(np.float32)

gpr = GPRegressor(
    VariationalRegressionModule,
    module__inducing_points=torch.from_numpy(X[:10]),
    criterion=gpytorch.mlls.VariationalELBO,
    criterion__num_data=50,
    verbose=0,
).initialize()

assert isinstance(gpr.module_.variational_strategy, gpytorch.variational.VariationalStrategy)
pickle.dumps(gpr.module_.variational_strategy)  # works

gpr.fit(X, y)  # <= fitting is necessary
pickle.dumps(gpr.module_.variational_strategy)  # now it fails
```

** Stack trace/error message **
```
_pickle.PicklingError: Can't pickle <function CholLinearOperator.to_dense at 0x79e1909fe840>: it's not the same object as linear_operator.operators.chol_linear_operator.CholLinearOperator.to_dense
```

## Expected Behavior

With v1.14, this snippet works.

## System information

**Please complete the following information:**
- 1.15.0, 1.15.1
- 2.10.0
- Linux

kayween:
### Context

This PR aims to fix https://github.com/cornellius-gp/gpytorch/issues/2717

The recent changes in GPyTorch v1.15 make variational models not pickle-able anymore. After digging around, it seems that the root cause lies in linear operators, as linear operators are not pickle-able in the first place. See [here](https://github.com/cornellius-gp/gpytorch/issues/2717#issuecomment-3911468697) for a code snippet.

### This PR
Claude suggested that the issue is caused by the `@cache` decorator. By default, the cache uses the original **unwrapped** function in the dictionary key; see [here](https://github.com/cornellius-gp/linear_operator/blob/34da7b6b194a2af9f3594a898e5c844d87a5e79c/linear_operator/utils/memoize.py#L57). But pickle looks up the function by the qualified name, which points to the **wrapped** function. This eventually causes a check failure in pickle dumps. This PR fixes this by using the qualified name (a string) in the dictionary key.

~~P.S. There is still a test failure in the kernel linear operator, which is likely caused by lambda functions.~~

### DIFF
--- a/linear_operator/operators/kernel_linear_operator.py
+++ b/linear_operator/operators/kernel_linear_operator.py
@@ -14,6 +14,10 @@ from linear_operator.utils.getitem import _noop_index, IndexType
 from linear_operator.utils.memoize import cached
 
 
+def _default_num_nonbatch_dimensions():
+    return 2
+
+
 def _x_getitem(x, batch_indices, data_index):
     """
     Helper function to compute x[*batch_indices, data_index, :] in an efficient way.
@@ -142,9 +146,9 @@ class KernelLinearOperator(LinearOperator):
     ):
         # Change num_nonbatch_dimensions into a default dict
         if num_nonbatch_dimensions is None:
-            num_nonbatch_dimensions = defaultdict(lambda: 2)
+            num_nonbatch_dimensions = defaultdict(_default_num_nonbatch_dimensions)
         else:
-            num_nonbatch_dimensions = defaultdict(lambda: 2, **num_nonbatch_dimensions)
+            num_nonbatch_dimensions = defaultdict(_default_num_nonbatch_dimensions, **num_nonbatch_dimensions)
 
         # Divide params into tensors and non-tensors
         tensor_params = dict()
--- a/linear_operator/test/linear_operator_test_case.py
+++ b/linear_operator/test/linear_operator_test_case.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 
 import logging
 import math
+import pickle
 import traceback
 from abc import abstractmethod
 from itertools import combinations, product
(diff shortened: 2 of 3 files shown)


gpleiss (approved):
LGTM!

PR #126, opened 2026-03-03
Fix `add_low_rank` to only compute roots when cached roots exist

SebastianAment:
This fixes a numerical instability bug where add_low_rank would speculatively compute root decompositions even when no cached roots existed. This caused SVD failures on ill-conditioned matrices (e.g., when using RBF + LinearKernel).

Changes:
- Change early exit condition from `not generate_roots and not has_roots` to `not (generate_roots and has_roots)` - exit early if either `generate_roots=False` or no cached roots exist
- Pass `generate_roots` parameter through in `RootLinearOperator.add_low_rank`, which was being silently ignored
- Add regression test to verify `root_decomposition` is not called when no roots are cached


gpleiss (approved):
LGTM

PR #127, opened 2026-03-04
Replace deprecated torch APIs with modern torch.linalg equivalents

saitcakmak:
torch.inverse, torch.pinverse, and torch.norm have been deprecated since PyTorch 1.9. This updates all usage to their modern replacements and, critically, registers torch.linalg.inv for __torch_function__ dispatch so that torch.linalg.inv(linear_op) works correctly.

Changes:
- Register torch.linalg.inv alongside torch.inverse for LinearOperator dispatch (fixes torch.linalg.inv not working on LinearOperators)
- Replace torch.pinverse() with torch.linalg.pinv()
- Replace torch.norm() with torch.linalg.vector_norm() (source files) and torch.linalg.norm() (test files)
- Update stale comments referencing torch.cholesky, torch.solve, torch.symeig, and torch.eig to their modern equivalents

### DIFF
--- a/examples/LinearOperator_demo.ipynb
+++ b/examples/LinearOperator_demo.ipynb
@@ -220,7 +220,7 @@
    "source": [
     "#### Eigendecomposition\n",
     "\n",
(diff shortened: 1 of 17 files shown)


Balandat (approved):
Great, thanks!

PR #128, opened 2026-03-04
Fix deprecated APIs, latent bugs, and code quality issues

saitcakmak:
### Summary

  Follow-up to #127. Audited the codebase for remaining deprecated API usage, latent bugs, and code quality issues.

###   Deprecated API Replacements

  - torch.cholesky_solve → torch.linalg.solve_triangular (low_rank_root_added_diag_linear_operator.py): torch.cholesky_solve has been deprecated since PyTorch 1.9.
  Replaced with two chained torch.linalg.solve_triangular calls (forward solve with L, then back-substitution with L^T).
  - torch.dsmm → torch.mm (sparse.py, 3 call sites): torch.dsmm is an undocumented/legacy sparse-dense matrix multiply. torch.mm handles sparse inputs natively.
  - Old-style sparse tensor constructors → torch.sparse_coo_tensor (sparse.py, 3 call sites): Replaced getattr(torch.sparse, "FloatTensor")(...) patterns with the
  modern torch.sparse_coo_tensor().
  - bool_compat workaround → torch.bool (deprecation.py, linear_cg.py, cat_linear_operator.py): Removed workaround for PyTorch issue #21113 (resolved since PyTorch
  1.2). Replaced all bool_compat with torch.bool directly.

###   Bug Fixes

  - Missing f-string prefix (cat_linear_operator.py:297): Error message "Unexpected index type {cat_dim_indices.__class__.__name__}..." was missing the f prefix —
  the {...} was printed literally instead of interpolated.
  - _batch_move_memo naming and tuple mismatch (batch_repeat_linear_operator.py): Two bugs: (1) The setter used self.__batch_move_memo (name-mangled) but hasattr
  checked _batch_move_memo — the memo was never retrieved. (2) The setter stored a 3-tuple but the getter unpacked only 2 values, which would raise ValueError if
  ever hit. Fixed by using _batch_move_memo consistently and unpacking all 3 values.
  - Duplicate _diagonal() call (triangular_linear_operator.py:197): self._diagonal() was called twice — once stored in diag, then called again for logdet_term. Used
   the already-computed diag instead.

###   Code Quality

  - Removed duplicate create_linear_op() call in test_add_low_rank
  - Updated stale torch.solve comment → torch.linalg.solve
  - class StochasticLQ(object) → class StochasticLQ:
  - Removed unnecessary .contiguous() calls before .expand() and summation

### DIFF
--- a/linear_operator/operators/_linear_operator.py
+++ b/linear_operator/operators/_linear_operator.py
@@ -80,7 +80,7 @@ def _implements_second_arg(torch_function: Callable) -> Callable:
     where the first argument of the function is a torch.Tensor and the
     second argument is a LinearOperator
 
-    Examples of this include :meth:`torch.cholesky_solve`, `torch.solve`, or `torch.matmul`.
+    Examples of this include :meth:`torch.cholesky_solve`, `torch.linalg.solve`, or `torch.matmul`.
     """
 
     @functools.wraps(torch_function)
--- a/linear_operator/operators/_linear_operator.py
+++ b/linear_operator/operators/_linear_operator.py
@@ -80,7 +80,7 @@ def _implements_second_arg(torch_function: Callable) -> Callable:
     where the first argument of the function is a torch.Tensor and the
     second argument is a LinearOperator
 
-    Examples of this include :meth:`torch.cholesky_solve`, `torch.linalg.solve`, or `torch.matmul`.
+    Examples of this include `torch.linalg.solve` or `torch.matmul`.
     """
 
     @functools.wraps(torch_function)
--- a/linear_operator/operators/kronecker_product_linear_operator.py
+++ b/linear_operator/operators/kronecker_product_linear_operator.py
@@ -35,7 +35,7 @@ def _matmul(linear_ops, kp_shape, rhs):
     output_shape = _matmul_broadcast_shape(kp_shape, rhs.shape)
     output_batch_shape = output_shape[:-2]
 
-    res = rhs.contiguous().expand(*output_batch_shape, *rhs.shape[-2:])
+    res = rhs.expand(*output_batch_shape, *rhs.shape[-2:])
     num_cols = rhs.size(-1)
     for linear_op in linear_ops:
         res = res.view(*output_batch_shape, linear_op.size(-1), -1)
@@ -50,7 +50,7 @@ def _t_matmul(linear_ops, kp_shape, rhs):
     output_shape = _matmul_broadcast_shape(kp_t_shape, rhs.shape)
     output_batch_shape = torch.Size(output_shape[:-2])
 
-    res = rhs.contiguous().expand(*output_batch_shape, *rhs.shape[-2:])
+    res = rhs.expand(*output_batch_shape, *rhs.shape[-2:])
     num_cols = rhs.size(-1)
     for linear_op in linear_ops:
         res = res.view(*output_batch_shape, linear_op.size(-2), -1)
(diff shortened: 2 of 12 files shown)


linear_operator/operators/kronecker_product_linear_operator.py
```diff
@@ -50,7 +50,7 @@ def _t_matmul(linear_ops, kp_shape, rhs):
     output_shape = _matmul_broadcast_shape(kp_t_shape, rhs.shape)
     output_batch_shape = torch.Size(output_shape[:-2])
 
-    res = rhs.contiguous().expand(*output_batch_shape, *rhs.shape[-2:])
+    res = rhs.expand(*output_batch_shape, *rhs.shape[-2:])
```
Balandat, line 53: Are we sure these are not necessary? I've seen some pretty weird behavior with `contiguous()` esp. when things like `LinearOperator`s are involved. This seems ok but I had thought that in other situations as well...

linear_operator/operators/_linear_operator.py
Balandat: cholesky_solve should go if deprecated and you're removing it below

linear_operator/utils/deprecation.py
```diff
@@ -3,15 +3,10 @@
...
-from unittest.mock import MagicMock
 
 import torch
 
-# TODO: Use bool instead of uint8 dtype once pytorch #21113 is in stable release
-if isinstance(torch, MagicMock):
-    bool_compat = torch.uint8
-else:
-    bool_compat = (torch.ones(1) > 0).dtype
+bool_compat = torch.bool
```
Balandat: is this used anywhere? Why not kill it?

linear_operator/utils/sparse.py
```diff
@@ -114,21 +111,21 @@ def bdsmm(sparse, dense):
...
         dense = dense.view(batch_size, num_rows, num_cols)
-        res = torch.dsmm(sparse, dense.transpose(0, 1).reshape(-1, batch_size * num_cols))
+        res = torch.mm(sparse, dense.transpose(0, 1).reshape(-1, batch_size * num_cols))
         res = res.view(-1, batch_size, num_cols)
         res = res.transpose(0, 1).reshape(*batch_shape, -1, num_cols)
         return res
 
     else:
-        return torch.dsmm(sparse, dense)
+        return torch.mm(sparse, dense)
```
Balandat: is this equivalent to `torch.sparse.mm`? Note that `torch.mm` has the following warning that `torch.sparse.mm` does not have:

> Sparse support is a beta feature and some layout(s)/dtype/device combinations may not be supported, or may not have autograd support. If you notice missing functionality please open a feature request.

Balandat (changes requested):
Thanks!

linear_operator/operators/kronecker_product_linear_operator.py
saitcakmak: I questioned these as well and the response was that expand operation will result in non-contiguous view anyways, so you don't benefit from the prior contiguous call here. If we need a contiguous view, I think the right move would be to call contiguous after expand, or use reshape instead

linear_operator/utils/deprecation.py
saitcakmak: yeah, I was gonna tell it to remove it

linear_operator/utils/sparse.py
saitcakmak: We can use torch.sparse.mm as well. That was the other suggestion

Balandat (approved):
thanks

PR #132, opened 2026-05-20
Avoid torch.sparse_coo_tensor DeprecationWarning

### LINKED ISSUE #131: [Bug] Explicit `check_invariants` needed
# 🐛 Bug

Since PyTorch 2.1, `torch.sparse_coo_tensor` emits a `UserWarning` when `check_invariants` is not passed explicitly.

## To reproduce

```python
import warnings

import torch
from linear_operator.operators import InterpolatedLinearOperator

warnings.filterwarnings(
    "error",
    message="Sparse invariant checks are implicitly disabled",
    category=UserWarning,
)

base = torch.eye(3)
indices = torch.tensor([[0], [1], [2]])
op = InterpolatedLinearOperator(base, indices, indices)
op.to_dense()
```

**Stack trace/error message**
```
Sparse invariant checks are implicitly disabled. Memory errors (e.g. SEGFAULT) will occur when operating on a sparse tensor which violates the invariants, but checks incur performance overhead. To silence this warning, explicitly opt in or out.
```

## Expected Behavior

`check_invariants` needs to be explicitly passed to all calls of `torch.sparse_coo_tensor`.

## System information
- LinearOperator Version: 0.6.1
- PyTorch Version: 2.11.0
- Computer OS: macOS

AdrianSosic:
Fixes #131 by explicitly passing `check_invariants=False` to all calls of `torch.sparse_coo_tensor`, staying in line with the previous behavior but avoiding the `DeprecationWarning`.

### DIFF
--- a/linear_operator/utils/sparse.py
+++ b/linear_operator/utils/sparse.py
@@ -60,7 +60,12 @@ def make_sparse_from_indices_and_values(interp_indices, interp_values, num_rows)
     # Make the sparse tensor
     interp_size = torch.Size((*batch_shape, num_rows, n_target_points))
     res = torch.sparse_coo_tensor(
-        index_tensor, value_tensor, interp_size, dtype=value_tensor.dtype, device=value_tensor.device
+        index_tensor,
+        value_tensor,
+        interp_size,
+        dtype=value_tensor.dtype,
+        device=value_tensor.device,
+        check_invariants=False,
     )
 
     # Wrap things as a variable, if necessary
@@ -108,6 +113,7 @@ def bdsmm(sparse, dense):
             torch.Size((batch_size * num_rows, batch_size * num_cols)),
             dtype=sparse._values().dtype,
             device=sparse._values().device,
+            check_invariants=False,
         )
 
         dense_2d = dense.reshape(batch_size * num_cols, -1)
@@ -134,7 +140,7 @@ def sparse_eye(size):
     """
     indices = torch.arange(0, size).long().unsqueeze(0).expand(2, size)
     values = torch.tensor(1.0).expand(size)
-    return torch.sparse_coo_tensor(indices, values, torch.Size([size, size]))
+    return torch.sparse_coo_tensor(indices, values, torch.Size([size, size]), check_invariants=False)
 
 
 def sparse_getitem(sparse, idxs):
@@ -203,7 +209,9 @@ def sparse_getitem(sparse, idxs):
             case _:
                 raise RuntimeError("Unknown index type")
 
-    return torch.sparse_coo_tensor(indices, values, torch.Size(size), dtype=values.dtype, device=values.device)
+    return torch.sparse_coo_tensor(
+        indices, values, torch.Size(size), dtype=values.dtype, device=values.devicei, check_invariants=False
+    )
 
 
 def sparse_repeat(sparse, *repeat_sizes):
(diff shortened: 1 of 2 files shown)


linear_operator/utils/sparse.py
```diff
@@ -203,7 +209,9 @@ def sparse_getitem(sparse, idxs):
             case _:
                 raise RuntimeError("Unknown index type")
 
-    return torch.sparse_coo_tensor(indices, values, torch.Size(size), dtype=values.dtype, device=values.device)
+    return torch.sparse_coo_tensor(
+        indices, values, torch.Size(size), dtype=values.dtype, device=values.devicei, check_invariants=False
```
kayween: `device=values.devicei` this seems to be a typo?

kayween:
hmm it seems that setting the flag `check_invariants` explicitly does not suppress the warnings. Let me see if there's an alternative way to fix the warning.

linear_operator/utils/sparse.py
AdrianSosic: yes, thx for the fix

AdrianSosic:
> hmm it seems that setting the flag `check_invariants` explicitly does not suppress the warnings. Let me see if there's an alternative way to fix the warning.

Interesting indeed! The torch docs are not very explicit about it, but if I understand the situation right, the problem is not the missing arguments but the fact that the *global* state for the check has not been set. But setting that is no real alternative since it would also affect other packages consuming `gpytorch`, so the only real option I can see at the moment is to wrap each call in `with torch.sparse.check_sparse_tensor_invariants(False)` context, which seems to work. Is that acceptable?

The second question would be: should the checks still remain active for the tests?

kayween:
I've opened up an issue on the pytorch side to see if the warning can be suppressed by the flag https://github.com/pytorch/pytorch/issues/184610

Now responding to your questions:
1. It sounds good to me to wrap `sparse_coo_tensor` inside the context manager.
1. I am fine either way. It's probably easier to just disable the check in the tests as well, because I'm not sure how to turn on the check in tests if the functional code already disables the check.

AdrianSosic:
> I've opened up an issue on the pytorch side to see if the warning can be suppressed by the flag [pytorch/pytorch#184610](https://github.com/pytorch/pytorch/issues/184610)
> 
> Now responding to your questions:
> 
> 1. It sounds good to me to wrap `sparse_coo_tensor` inside the context manager.
> 2. I am fine either way. It's probably easier to just disable the check in the tests as well, because I'm not sure how to turn on the check in tests if the functional code already disables the check.

@kayween: Thanks, very good idea! This is actually good news because since the problem is really on the torch side, we can simply keep the existing changes and they will start to work once the problem is fixed upstream, right? So no need for the context manager then!

RE tests: The only way I see is to make this happen is by using some trigger that is set during the tests, e.g. an environment variable that alters the flag (or in case of the decorator, the behavior of it). One could potentially weave this into the `settings` mechanism somehow, I guess?

kayween:
@AdrianSosic We can keep the changes here. Let's just wait for https://github.com/pytorch/pytorch/pull/183256 to land on the pytorch side.

AdrianSosic:
@kayween: I think the upstream change has been completed. Shall we proceed with the PR then?

kayween:
@AdrianSosic Yeah, I will merge it after testing the PR with the latest pytorch build.

kayween (approved):

